PR #3, opened 2023-09-09
feat: Add person models

jennifer-richards:
Draft PR notes: new work starts with c7fa4ba291960dcc50806d827265343ff4b978c2, earlier is from #2.

#### Concept
This brings over `RpcPerson` and person-related models from the `feat/rfc` branch and proposes a mechanism for linking datatracker `Person` instances to models in this application that will facilitate handling `Person` merges.

I've suggested that we add a `DatatrackerPerson` model that shadows the datatracker's `Person`. A many-to-one relationship between `DatatrackerPerson` and `RpcPerson` is used to connect the RPC's concept of a real person with one or more datatracker accounts. In this framework, we will replace `RpcPerson` in most relationships with a `DatatrackerPerson`. These will capture the `Person` understood to be connected _at the time the record was created_.

With this, I believe a datatracker `Person` merge can be reflected by changing the foreign key from the disfavored `DatatrackerPerson` to point to the `RpcPerson` of the one merged into.

When querying objects pertaining to a `DatatrackerPerson`, the process is then something like:
```python
my_rfcs_to_be = RfcToBe.objects.filter(author__person__rpcperson=current_user__person__rpcperson)
```

#### Nuts and bolts
The `DatatrackerPerson` model could perhaps be merged into the `User` model, but I slightly prefer this because it makes clearer the meaning of a `User` without a `DatatrackerPerson`.

I've allowed `DatatrackerPerson` not to have an `RpcPerson`. I don't know if there's a need for this.


jennifer-richards:
Going to replace with a reworked alternative rather than try to fix this up in place.

PR #4, opened 2023-09-14
chore: Reorganize apps and start adding models

jennifer-richards:


rpcauth/backends.py
```diff
@@ -0,0 +1,65 @@
...
+        subject_id = claims["sub"]
+        # A DatatrackerPerson may exist without a user so use get_or_create()
+        datatracker_person, _ = DatatrackerPerson.objects.get_or_create(
+            subject_id=subject_id,
+            plain_name=claims.get("name", ""),
+        )
+        try:
+            # Rely on the OneToOne field to avoid creating duplicate Users for a DatatrackerPerson
+            new_user = self.UserModel.objects.create(
+                username=f"dt-user-{subject_id}",
```
rjsparks, line 31: Consider dt-person, so future us aren't nudged into looking in the datatracker user tables.

PR #9, opened 2023-09-19
fix: adjust rpc_person api to match model changes

rjsparks:
See the imbedded question.

@NGPixel asked if the client would ever touch the datatracker directly - this might be a place to ask.

If it _was_ setup to touch the datatracker directly, would we need to stash things like plain_name in the rpc database?


rjsparks:
To be clear, we could provide an endpoint to the client from the datatracker that gave people's names given a set of pks.

jennifer-richards (approved):

PR #12, opened 2023-09-27
feat: Use bulk lookup of Person name (alt)

jennifer-richards:
This does the same as #11 but uses the alternative API endpoint suggested in https://github.com/ietf-tools/datatracker/pull/6392 to avoid having to convert an integer `Person.id` into a string.

Only one of these should be merged, so I'm making them both draft.


jennifer-richards:
Closing in favor of #11

PR #13, opened 2023-09-27
refactor: Rationalize the API usage a bit

jennifer-richards:
Mostly just gets a few things out of the `datatracker.models` file where they'd been stashed for convenience.

@NGPixel- could you check in particular the 3rd commit, which puts all the config for the code generator call into one file (the previously ignored `openapitools.json` file). It seems to work for me both with vscode and the old-fashioned way but I'm not familiar enough with vscode to be sure I didn't break it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,3 @@ __pycache__
 
 /rpcapi.yaml
 /openapi/rpcapi_client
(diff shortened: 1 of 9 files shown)


NGPixel (approved):
Worked just fine in vscode

PR #58, opened 2023-11-02
feat: Hide mgmt items from non-managers

jennifer-richards:
Among the changes, hides the "Add team member" button. Not sure if we want to entirely hide the "Team" page or not. It's currently titled "Manage Team Members," but maybe editors should get a "View Team Members"?

### DIFF
--- a/client/components/SidebarNav.vue
+++ b/client/components/SidebarNav.vue
@@ -73,8 +73,8 @@
               </li>
             </ul>
           </li>
-          <li>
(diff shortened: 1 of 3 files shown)


rjsparks:
> maybe editors should get a "View Team Members"

lets work that out with Jean

I think that everyone should get to create labels though.

jennifer-richards:
Added back "Manage labels" for everyone. We may want to consider whether we need to protect against someone accidentally renaming or removing a core label (though we don't yet support removing labels)

PR #72, opened 2023-12-04
feat: More history work

jennifer-richards:
This reorganizes the `RfcToBe` history serialization. It introduces a reusable serializer that can be applied to other models that need to display histories, returns user metadata for the "by" field, and adds tracking of label history.

The current implementation is fairly intensive of both SQL queries (for label histories) and datatracker API calls (for user name lookups) - I'm sure this can be improved, but I'm leaving it raw for now.

It's apparent from playing with this that we'll want to be sure the RPC personnel are trained to be thoughtful about changing label names. Histories and statistics will get to be a mess if labels are repurposed.


rjsparks:
@ajeanmahoney - see the comment about labels above for discussion

ajeanmahoney:
> be sure the RPC personnel are trained to be thoughtful about changing label names. Histories and statistics will get to be a mess if labels are repurposed.

Ack. Understood

PR #88, opened 2023-12-28
feat: import information from production rfc editor database.

rjsparks:


rfced/models.py
```diff
@@ -0,0 +1,507 @@
+# This is an auto-generated Django model module.
+# You'll have to do the following manually to clean this up:
+#   * Rearrange models' order
+#   * Make sure each model has one field with primary_key=True
+#   * Make sure each ForeignKey and OneToOneField has `on_delete` set to the desired behavior
+#   * Remove `managed = False` lines if you wish to allow Django to create, modify, and delete the table
+# Feel free to rename the models, but don't rename db_table values or field names.
```
rjsparks, line 7: I've done _none_ of these things. The models are only useful for building queries that don't involve joining tables. I'm expecting we will do the joins in python (or in raw sql) when we get to them, rather than try to mutate the database to the point that django will more fully work with it.

PR #149, opened 2024-08-06
fix: DocumentTable sort column width wobble.

### LINKED ISSUE #49: nit: table component changes column widths when sorting by different columns
Sort different columns on the queue tabs and watch the table jump around

holloway:
Adds a placeholder icon to each column so that each column has the same width regardless of whether it's sorted.

fixes #49

question: does this need tests? I'm guessing not.

We should probably consider changing the `<a>`s to be `<button>`s and use [`aria-sort`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort) for a11y reasons.

### DIFF
--- a/client/components/DocumentTable.vue
+++ b/client/components/DocumentTable.vue
@@ -14,8 +14,15 @@
               <span>{{ col.label }}</span>
               <template v-if="state.sortField === col.field">
                 <Icon v-if="state.sortDirection === 'asc'" name="uil:arrow-up" class="text-lg -mt-0.5" />
-                <Icon v-else name="uil:arrow-down" class="text-lg -mt-0.5" />
+                <Icon v-else-if="state.sortDirection === 'desc'" name="uil:arrow-down" class="text-lg -mt-0.5" />
               </template>
+              <template v-else>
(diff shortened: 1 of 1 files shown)


rjsparks:
> We should probably consider changing the <a>s to be <button>s and use [aria-sort](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort) for a11y reasons.

Is there any reason _not_ to make this change?

rjsparks (approved):

jennifer-richards (approved):

jennifer-richards:
> question: does this need tests? I'm guessing not.
> 

Not yet. At some point we need to start writing tests but so far we've been taking a rapid prototyping approach here rather than TDD.

> We should probably consider changing the `<a>`s to be `<button>`s and use [`aria-sort`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort) for a11y reasons.

Sounds good to me

holloway:
@rjsparks there's no downside, so yes we should do this

PR #151, opened 2024-08-07
feat: Assignee people

holloway:
The Assignee column now looks like this...

Screenshot from 2024-08-07 13-25-03

Is that the intended formatting? Does this need tests?


rjsparks (approved):

jennifer-richards (approved):
It looks ok to me.

We could perhaps restyle it a bit (showing a person with multiple assignments as a single name with multiple badges jumps out at me as something to try), but that can be a future exploration / enhancement.

PR #165, opened 2024-08-16
feat: assignment as dropdown (non drag)

holloway:
Allows assignments by dropdown rather than drag-n-drop (because the dragable areas were visibly too large to be easy to use)

It looks like this

Screenshot from 2024-08-21 09-09-18

***NOTE WELL*** : The above data looks like real drafts and real people, but the dates and assignments are test data, not real data. This doesn't reflect what's actually happening in the RPC right now at all.


ajeanmahoney:
Is the order alphabetical by first name? If someone could complete the assignment earlier, would they be listed first?

rjsparks:
We can make the order by "best match". For now, that could be "earliest date".

holloway:
I'll change the ordering.

What do you think of the UX in general? Is it usable? Any feedback?

If it's ok then I'll tidy up this PR and ask for a code review.

ajeanmahoney:
Inactive editors (i.e., Jon Postel, Joyce Reynolds) should not be listed in the drop-down menu. If the editor already is working on an assignment, the description should say that (e.g., "Currently assigned draft-foo-06, 86 pages"), thanks!

rjsparks:
removing inactive editors will come in with different work.

@holloway - tidy this up as suggested. I'll make another issue for adding information about current assignments to the dropdowns.

holloway:
@rjsparks done and screenshot updated on this PR.

fixes #168

jennifer-richards (approved):
lgtm. Hope you don't mind, but I removed a few prototype components that were left unused by your refactoring.

jennifer-richards (approved):

PR #172, opened 2024-08-23
fix: remove exports / checks that broke main

jennifer-richards:


rjsparks (approved):
Should follow up on the test for documentId type.

jennifer-richards:
> Should follow up on the test for documentId type.

I'm pretty sure it's either because it's adding a permanent failure in a spot where the value is only temporarily missing or because the type isn't actually "number" because of some proxy involved in the async magic.

PR #173, opened 2024-08-23
feat: mark names used, provided unknown names

rjsparks:


jennifer-richards (approved):
Noting for posterity that @rjsparks and I discussed the so-far unused `used` field this adds. We agreed to keep it for now and can remove it later if it turns out not to be needed.

jennifer-richards (approved):

PR #175, opened 2024-08-27
chore: TypeScript

holloway:
...continuing to add TS


client/package.json
```diff
@@ -27,11 +28,12 @@
     "eslint-config-standard": "17.1.0",
     "eslint-plugin-nuxt": "4.0.0",
     "eslint-plugin-vue": "9.24.0",
-    "nuxt": "3.11.1",
+    "nuxt": "3.12.3",
```
jennifer-richards, line 31: Anything significant in this upgrade that would warrant making it a separate PR?

client/pages/docs/import.vue
```diff
@@ -117,7 +117,7 @@ const timeToDeadline = computed(() => {
     if (state.deadline) {
       const dt = DateTime.fromISO(state.deadline).diff(today, 'days')
       return humanizeDuration(
-        dt,
+        dt.toMillis(),
```
jennifer-richards, line 120: 🎉

jennifer-richards (commented):
Looks ok to me - one question about the nuxt upgrade. Also, the vue examples hint but don't outright say that dropping the `const props =...` is ok when `props` is not used in the `<script>`. Did you test to confirm this? (e.g., check that badges come out in different colors; I imagine that if there were a problem with the props, it'd be really obvious, just wanted to ask)

Just commenting - I'll let Nick handle actually approving.

holloway:
@jennifer-richards I tried to test it but I've had problems testing these PRs in general so I'm not confident. I've been copying `/client/` (except `node_modules`) into rfced-xfer's `/client/` to test, doing an `npm install`, but I've had random errors. I've tried cleaning Nuxt cache etc. but that doesn't seem to be the cause.

How are you testing?

I've just restored the `const props=` and fixed an import mistake.

client/package.json
holloway: This was TS-related .. see https://github.com/nuxt/nuxt/discussions/25530#discussioncomment-9951962

holloway: I'm happy to split up this PR though if you'd like

jennifer-richards: No, seems on-topic enough

jennifer-richards:
I checked this out and it runs - taking the `const props =...` out of `ConfirmDialog.vue` seems to be ok, so I don't object to taking those out back if you prefer it that way.

For testing, I run a datatracker instance with `feat/rpc-api` and run the `rpc` project directly. Unless I'm specifically looking at imported data, I use the data created via `./manage.py create_rpc_demo` to exercise the UI. As we add features, we should probably update the `create_rpc_demo.py` script to fill in enough of a dev database to do a full demo. At some point I imagine we'll create a pre-built dev database image but I don't think we have a specific plan yet.

NGPixel (approved):

jennifer-richards:
Going to merge this as it is - if you'd like to clean out the (it seems to me) unnecessary `const props...` bits, feel free to do so going forward.

PR #176, opened 2024-08-29
feat: currentTime composable

jennifer-richards:
Adds a `currentTime` composable to provide a reactive single source of truth for the current time


client/composables/useCurrentTime.ts
```diff
@@ -0,0 +1,25 @@
+import { ref, onBeforeMount, onUnmounted, readonly } from 'vue'
+import { DateTime } from 'luxon'
+
+const getNowUtc = () => DateTime.now().setZone('utc')
+const currentTime = ref(getNowUtc())
+let interval: number = 0
```
holloway, line 6: If the inferred typing is the same as the explicit typing then I reckon it's to omit the type declaration `number`.

This code is good so no changes are requested, but `undefined | number` might be clearer than using `0` as the unset state. Another approach rather than `undefined | number` would be `undefined | ReturnType<typeof setInterval>` which is the exact same typing, but it indicates that it's whatever the value returned by `setInterval(...)` is.

```suggestion
let interval: undefined | ReturnType<typeof setInterval>
```

holloway (approved):

client/composables/useCurrentTime.ts
jennifer-richards: I've adopted the `ReturnType` suggestion. It revealed potential confusion about the return type. My dev environment thinks `ReturnType<typeof setInterval>` is different from what `window.setInterval()` returns. It doesn't help to use `ReturnType<typeof window.setInterval>`, either. This seems to be a distinction between NodeJS and browser side implementations. Everything seems happy if I drop the `window` specification and it seems to work, so I'm doing that.

Also, I'm using `null` rather than `undefined` as the default case. I think that is a compromise between understanding why 0 is a magic number and my own squickiness with setting a variable back to `undefined` or otherwise working around it.

PR #178, opened 2024-08-30
feat: real "draft import" functionality

jennifer-richards:
This eliminates the factory-generated and demo data that were in the initial "import draft" workflow. This is by no means a "finished" page, but I think it's a good checkpoint for the basic functionality.

With the caveat that the data shown on this image are based on real Datatracker data but have no connection to actual RPC / RFC Editor work, this is what the import UI currently looks like:
image

The data in the box at the top come from the datatracker. The listboxes below are initialized to match or set to "Select..." for the boilerplate. Labels can be assigned. Once the listboxes are filled in, the "Save" button in the corner goes live and an `RfcToBe` will be created.

@holloway, I've tried to type-scriptify things here, can you please look that over? This overlaps somewhat with your most recent PR.

There's a PR also needed to `feat/rpc-api` on the datatracker side, which I'll create shortly.


client/pages/docs/import.vue
```diff
@@ -83,41 +245,46 @@
...
 
-const state = reactive({
+type State = {
+  boilerplate: TlpBoilerplateChoiceName | null,
+  sourceFormat: SourceFormatName | null,
+  stdLevel: StdLevelName | null,
+  stream: StreamName | null,
+  deadline: string | null,
+  labels: number[]
+}
```
holloway, line 263: typing looks good :+1:

client/pages/queue/[section].vue
```diff
@@ -84,8 +84,8 @@ const tabs: Tab[] = [
 const deadlineCol = {
   key: 'deadline',
   label: 'Deadline',
-  field: 'deadline',
-  format: (val: any) => val ? DateTime.fromISO(val).toLocaleString(DateTime.DATE_MED_WITH_WEEKDAY) : '',
+  field: 'externalDeadline',
+  format: (val: any) => val ? DateTime.fromJSDate(val as Date).toLocaleString(DateTime.DATE_MED_WITH_WEEKDAY) : '',
```
Comment (holloway) [inline client/pages/queue/[section].vue:88]: `as` casting is a teensy bit risky as it can hide errors... perhaps something like this?
 
```suggestion
  format: (val: any) => val instanceof Date ? DateTime.fromJSDate(val).toLocaleString(DateTime.DATE_MED_WITH_WEEKDAY) : '',
```

datatracker/migrations/0003_document_intended_std_level_document_shepherd_and_more.py
```diff
@@ -0,0 +1,28 @@
...
+        migrations.AddField(
+            model_name='document',
+            name='intended_std_level',
+            field=models.CharField(blank=True, max_length=32),
+        ),
+        migrations.AddField(
+            model_name='document',
+            name='shepherd',
+            field=models.EmailField(blank=True, max_length=254),
+        ),
```
rjsparks, line 22: @ajeanmahoney - does RPC use the shepherd field? Would it change in the RPC app? Would it be sufficient to go look it up in the datatracker if needed instead of replicating it directly into the RPC app.

I ask because this is a place where the datatracker models need to change, and this is starting to leak the current design flaw from the datatracer into this app, and I'd rather not do that.

client/pages/queue/[section].vue
Comment (jennifer-richards) [reply client/pages/queue/[section].vue:88]: Hm, I'm worried this might still be hiding the error, perhaps even more deeply. I think an indication of some kind that the value is bad rather than absent would help.

It'd be nice if the type checks could go work across the component, but that seems like a tall order.

datatracker/migrations/0003_document_intended_std_level_document_shepherd_and_more.py
jennifer-richards: For this pr, I am happy not to add the field as I think we're likely to rework the document model anyway.

client/pages/queue/[section].vue
Comment (holloway) [reply client/pages/queue/[section].vue:88]: So there are 3 states: Date, not Date but truthy (error state), and undefined?

```suggestion
  format: (val: any) => val instanceof Date ? DateTime.fromJSDate(val).toLocaleString(DateTime.DATE_MED_WITH_WEEKDAY) : val ? 'Error: non-Date' : '',
```

I tried getting type checks working across the table component (ie applying column defs to row data, so that the `format` function would get a specific type) but retrofitting it was quite hard so I stopped and focused on other stuff. I could make a new table component with a similar API though.

Comment (jennifer-richards) [reply client/pages/queue/[section].vue:88]: A component that made this possible would be neat, but I wouldn't put it as a high priority. What you've got there with the "error" output looks fine to me.

rjsparks (approved):

PR #188, opened 2024-09-05
feat: add pylint

rjsparks:
This builds on #187, and should be evaluated after that is merged.

It brings in a copy of the datatracker's .pylintrc, which we should review carefully.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,408 @@
+[MASTER]
+
+# Specify a configuration file.
+#rcfile=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS,migrations
+
+# Add files or directories matching the regex patterns to the blacklist. The
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# List of plugins (as comma separated values of python modules names) to load,
+# usually to register additional checkers.
+load-plugins=pylint_django,pylint_common
+
+# Use multiple processes to speed up Pylint.
+jobs=1
+
+# Allow loading of arbitrary C extensions. Extensions are imported into the
+# active Python interpreter and may run arbitrary code.
+unsafe-load-any-extension=no
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code
+extension-pkg-whitelist=
+
+# Allow optimization of some AST trees. This will activate a peephole AST
+# optimizer, which will apply various small optimizations. For instance, it can
+# be used to obtain the result of joining multiple strings with the addition
+# operator. Joining a lot of strings can lead to a maximum recursion error in
+# Pylint and this flag can prevent that. It has one side effect, the resulting
+# AST will be different than the one from reality. This option is deprecated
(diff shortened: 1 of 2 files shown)


.pylintrc
```diff
@@ -0,0 +1,408 @@
...
+# Specify a configuration file.
+#rcfile=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS,migrations
```
rjsparks, line 12: CVS is vestigial
I'm surprised we're ignoring migrations.

.pylintrc
```diff
@@ -0,0 +1,408 @@
...
+
+# Add files or directories matching the regex patterns to the blacklist. The
```
rjsparks, line 14: blacklist

.pylintrc
```diff
@@ -0,0 +1,408 @@
...
+jobs=1
+
+# Allow loading of arbitrary C extensions. Extensions are imported into the
+# active Python interpreter and may run arbitrary code.
+unsafe-load-any-extension=no
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code
+extension-pkg-whitelist=
```
rjsparks, line 35: whitelist

We may not control this name, but I wonder if it's still a valid configuration name?

.pylintrc
```diff
@@ -0,0 +1,408 @@
...
+
+# Allow optimization of some AST trees. This will activate a peephole AST
+# optimizer, which will apply various small optimizations. For instance, it can
+# be used to obtain the result of joining multiple strings with the addition
+# operator. Joining a lot of strings can lead to a maximum recursion error in
+# Pylint and this flag can prevent that. It has one side effect, the resulting
+# AST will be different than the one from reality. This option is deprecated
+# and it will be removed in Pylint 2.0.
+optimize-ast=no
```
rjsparks, line 44: probably long gone - we are at pylint 3

.pylintrc
```diff
@@ -0,0 +1,408 @@
...
+# 3 compatible code, which means that the block might have code that exists
+# only in one or another interpreter, leading to false positives when analysed.
+analyse-fallback-blocks=no
+
+
+[EXCEPTIONS]
+
+# Exceptions that will emit a warning when being caught. Defaults to
+# "Exception"
+overgeneral-exceptions=builtins.Exception
```
rjsparks, line 408: anachronism?

.pylintrc
```diff
@@ -0,0 +1,408 @@
+[MASTER]
```
rjsparks, line 1: MASTER (new pylint generates MAIN)

.pylintrc
rjsparks: new pylint --generate-rcfile generates this same set

.pylintrc
rjsparks: from a newly generated rc:
```
# A comma-separated list of package or module names from where C extensions may
# be loaded. Extensions are loading into the active Python interpreter and may
# run arbitrary code. (This is an alternative name to extension-pkg-allow-list
# for backward compatibility.)
extension-pkg-whitelist=
```

.pylintrc
rjsparks: confirmed - no longer relevant

jennifer-richards (changes requested):
I think we should take the lessons from here and go with #189 instead

PR #189, opened 2024-09-05
feat: pylint with modern pylintrc

rjsparks:
Compare to #188.
Instead of fixing that old .pylintrc, start with a default generated one and change it into what we want.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -67,7 +67,8 @@
           "oderwat.indent-rainbow",
           "redhat.vscode-yaml",
           "spmeesseman.vscode-taskexplorer",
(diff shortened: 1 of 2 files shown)


jennifer-richards (commented):
This seems ok. It only explicitly sets up pylint for vscode, but that seems ok. It's present on the image by default and I haven't bothered committing pycharm configurations to the project so far...

There are a whole lot of violations:
```
Pylint found 98 errors, 249 warnings, 514 conventions, 291 refactors  in 96 files
```
Should we tweak the rules a bit now or commit this and adjust as we go? A lot of the complaints are obvious false alarms (migration filenames don't conform to snake case, SomeClass = apps.get_model(...) is mistaken for a variable name, etc)

PR #190, opened 2024-09-06
chore(deps): bump eslint/nuxt to latest; switch to neostandard

jennifer-richards:
Replaces eslint-config-standard with neostandard. To make that work, also migrates eslint to 9.9.1 and converts from `.eslintrc.js` to `eslint.config.mjs`. Resolving the dependencies required bumping nuxt up to a more recent version, so I've gone ahead and done that.

There will be more packages that require updates - I will sort those out and merge alongside this.

I've excluded a couple of generated code paths from being checked. That brings us down from about 3000 errors ~~to 15 errors and 2 warnings from `npx eslint`. I'm not planning to clean up that lint in this PR.~~ 0 after I failed not to clean up the code in this PR.

Thoughts? We could also ditch neostandard and use @holloway's approach with bare eslint.


NGPixel (approved):

jennifer-richards:
I couldn't resist - code is now free of eslint

PR #192, opened 2024-09-06
chore(deps): bump versions in package.json

jennifer-richards:
Updates versions of everything (or nearly everything, at least). Uses a `<21` version for `@types/node` to keep it in approximate sync with our nodejs version.


jennifer-richards:
Note that we are getting a warning related to `Vue3Snackbar` in the dev console. I think that's an upstream issue.

PR #204, opened 2024-09-07
fix: update create_rpc_demo.py

jennifer-richards:


rjsparks (approved):
We should discuss the future of the rpc demo constructors - relying on drafts that don't exist in the datatracker is probably going to run into trouble soon.

jennifer-richards:
> We should discuss the future of the rpc demo constructors - relying on drafts that don't exist in the datatracker is probably going to run into trouble soon.

Should be part of planning for getting back to a single fork. When not working on the xfer branch, the script is useful for reaching many parts of the UI.

PR #210, opened 2024-09-12
fix: remove field from labels col

### LINKED ISSUE #209: Clicking a document on the Queue page does not work
On any Queue page tab except the "Submissions" tab, clicking a doc name updates the URL bar in the browser but does not change to the document display page. It should.

An error appears in the javascript console.

As a workaround, reloading the page with the updated URL will show the desired page.

jennifer-richards:
Fixes #209 

This creates a problem because the `labels` field is a `Label` object and `<DocumentTable>` can't deal with it.

It was added as part of the typescript conversion and I'm not certain why, but I don't think it's needed functionally.


holloway (approved):
LGTM

PR #223, opened 2024-09-30
refactor: improve DocumentTable types

jennifer-richards:
This is an attempt to clean up some type complaints in the DocumentTable. It clears up all the complaints I see in DocumentTable.vue itself. Unfortunately the actual places the component is used are either seeing an `any` type or are confused because the `id` field is shown as possibly undefined due to limitations in the Django API generation. I'm hoping to clear those up eventually.

Also fixes a minor bug in the `labels` column where the ID was being displayed beside each label.

Best to look at the diffs with whitespace changes suppressed because some reformatting got mixed in - sorry about that.


jennifer-richards:
Going to draft - no need to look at this now, I am not sure it was doing what I thought it was doing. Will come back to it.

PR #225, opened 2024-10-01
chore: purge more tables

microamp:
Four additional tables to purge when the `purge` command runs:

| model                          | table                        |
|--------------------------------|------------------------------|
| `HistoricalRfcToBeLabel` \[1\] | `rpc_historicalrfctobelabel` |
| `HistoricalRfcToBe` \[1\]      | `rpc_historicalrfctobe`      |
| `HistoricalLabel` \[1\]        | `rpc_historicallabel`        |
| `RpcDocumentComment`           | `rpc_rpcdocumentcomment`     |

1.  managed by `django-simple-history`

### DIFF
--- a/rpc/management/commands/purge.py
+++ b/rpc/management/commands/purge.py
@@ -3,7 +3,18 @@
 from django.core.management.base import BaseCommand, CommandError, CommandParser
 
 from datatracker.models import DatatrackerPerson, Document
-from ...models import ActionHolder, Assignment, Cluster, Label, RfcToBe, RpcPerson
+from ...models import (
+    ActionHolder,
+    Assignment,
+    Cluster,
+    HistoricalLabel,  # type: ignore (managed by django-simple-history)
+    HistoricalRfcToBe,  # type: ignore (managed by django-simple-history)
+    HistoricalRfcToBeLabel,  # type: ignore (managed by django-simple-history)
+    Label,
+    RfcToBe,
+    RpcDocumentComment,
+    RpcPerson,
+)
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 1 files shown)


rpc/management/commands/purge.py
```diff
@@ -28,6 +39,10 @@ def handle(self, *args, **options):
 
         Assignment.objects.all().delete()
         ActionHolder.objects.all().delete()
+        HistoricalRfcToBeLabel.objects.all().delete()
+        HistoricalRfcToBe.objects.all().delete()
+        HistoricalLabel.objects.all().delete()
+        RpcDocumentComment.objects.all().delete()
```
rjsparks, line 45: I don't object to this, but I still think it should be happening automatically via database configuration (on_delete).
Or is there a reason to keep the history after you delete an object?

jennifer-richards: I'm not sure that works? Are we doing it somewhere? Looking at the rpc database,
```
rpctools=# \d rpc_historicalrfctobe
                                     Table "public.rpc_historicalrfctobe"
          Column          |           Type           | Collation | Nullable |             Default
[...]
Foreign-key constraints:
    "rpc_historicalrfctob_history_user_id_5e1aa891_fk_rpcauth_u" FOREIGN KEY (history_user_id) REFERENCES rpcauth_user(id) DEFERRABLE INITIALLY DEFERRED
```
i.e., there's no foreign key pointing to the `rpc_rfctobe` table that would trigger a cascade.

The history tracks deleted records, so in general at least the history tables have a different lifecycle from the records they describe.

jennifer-richards: From our offline discussion, for posterity the cascaded deletion is [the cascade_delete_history option](https://django-simple-history.readthedocs.io/en/latest/historical_model.html#deleting-historical-record)

jennifer-richards: For now let's be explicit about deleting history records.

PR #234, opened 2024-10-22
feat: k8s-ready configuration

jennifer-richards:
This deploys to a minikube cluster on my dev machine and runs well enough that I think it's time to move it to a more realistic environment.

It splits out development and production settings and deploys k8s in production mode with `DEBUG=False` and all that. FYI, I haven't yet gone through the full "production-ready" checklist for Django, so it's not expected that this is ready on that front. We may later want to add a staging or k8s-development mode with different settings, but this lets us differentiate local dev from k8s deployments as a place to start.


dev/build/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ghcr.io/ietf-tools/rpc-app-base:latest
+LABEL maintainer="IETF Tools Team <tools-discuss@ietf.org>"
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN groupadd -g 1000 rpc && \
+    useradd -c "Purple People Eater" -u 1000 -g rpc -m -s /bin/false purple
```
rjsparks, line 7: no way that won't come back as trouble later :P

jennifer-richards: Should be ok, he wouldn't eat you 'cause you're so tough

PR #273, opened 2025-04-30
feat: login route

holloway:
## feat
* Adds a `/login/` route asking the user to login. Hide all UI until authorised.

Screenshot_2025-05-01_15-09-26

### DIFF
--- a/client/components/HeaderNav.vue
+++ b/client/components/HeaderNav.vue
@@ -114,26 +114,22 @@
             </HeadlessMenuItems>
           </transition>
         </HeadlessMenu>
-        <div v-else class="-m-1.5 flex items-center p-1.5">
-
-
-          <DialogRoot :open="!userStore.authenticated">
-            <DialogPortal class="">
-              <DialogOverlay class="backdrop-blur-sm z-[100] bg-black/30 fixed inset-0" />
-              <DialogContent
-                class="fixed top-[50%] left-[50%] max-h-[85vh] w-[90vw] max-w-[450px] translate-x-[-50%] translate-y-[-50%] rounded-xl bg-white shadow-xl focus:outline-none z-[100] text-center px-3 py-4"
-              >
-                <DialogTitle class="mb-6 text-xs text-left font-bold text-gray-800 pl-3">
-                  You're not logged in
-                </DialogTitle>
-                <a href="/oidc/authenticate/" class="border bg-sky-900 px-4 py-2 text-white dark:text-neutral-300">
-                  <span class="font-semibold">Login via Datatracker</span>
-                  <Icon name="solar:alt-arrow-right-line-duotone" size="1.5em" class="ml-1" />
-                </a>
-              </DialogContent>
-            </DialogPortal>
-          </DialogRoot>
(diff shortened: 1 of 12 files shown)


client/components/HeaderNav.vue
```diff
@@ -114,21 +114,42 @@
...
         </HeadlessMenu>
-        <div v-else class="-m-1.5 flex items-center p-1.5">
-          <a href="/oidc/authenticate/" class="hidden lg:flex items-center text-sm dark:text-neutral-300">
-            <span class="font-semibold">Login</span>
-            <Icon name="solar:alt-arrow-right-line-duotone" size="1.5em" class="ml-1" />
-          </a>
-        </div>
+        <DialogRoot :open="!userStore.authenticated">
+          <DialogPortal>
+            <DialogOverlay class="backdrop-blur-sm z-[100] bg-black/30 fixed inset-0" />
```
jennifer-richards, line 119: The `z-[100]` could, I think, be `z-100` without affecting anything. Is there a reason to prefer the brackets, or is it just consistency?

jennifer-richards (approved):
Looks good. One question (just a question) inline.

rjsparks:
The current content behind the modal should be blank for a not logged in user. Will it be hard to make that change?

client/components/HeaderNav.vue
NGPixel: Brackets are usually for custom values but z-index handles all integers correctly without it as well. Both are valid so I would leave it to Matthew's preference.

NGPixel:
It would be best to add a dedicated "/login" page instead of a modal IMO. Then always redirect the user to it if not logged in from any other paths.

client/components/HeaderNav.vue
holloway: My take on it is that in general I think square bracket custom styles should be discouraged (or at least viewed with suspicion!) for CSS filesize reasons / consolidating unnecessary styles. They are often bespoke styles that aren't necessary, like using a custom box shadow vs a plain `shadow-xl` and the later should be preferred.

And arguably they should all be refactored into our own `Dialog.vue` file for reuse across the project, but I wasn't sure about commonalities yet.

The `z-*` styles will probably need rationalizing later.

holloway:
Sounds like we should have a video chat about UX. Might be more efficient than GH comments.

> The current content behind the modal should be blank for a not logged in user. Will it be hard to make that change?

No, but it would help to understand  they're not logged in so of course there won't be any API data blurred behind in the real thing (eg the blurry '23' wouldn't happen).

> It would be best to add a dedicated "/login" page instead of a modal IMO. Then always redirect the user to it if not logged in from any other paths.

As long as we can bounce back to the current page this sounds fine.

Personally I like the modal approach because the background has some per-route context that helps with multiple browser tabs open. I'm now talking about subtle UX context hints, and blurry ones at that, but it's more context than the `next` path param and I like them.

My understanding is that this also for the public to immediately tell them that it's not a public service (ie that it requires login)

Anyway -- might be worth a quick video chat, eh?

PR #274, opened 2025-05-01
feat: Login route

holloway:
## feat
* Adds a `/login/` route asking the user to login. Hide all UI until authorised.

Screenshot_2025-05-01_15-09-26

### DIFF
--- a/client/app.vue
+++ b/client/app.vue
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="!isOnLoginRoute && userStore.authenticated === undefined" class="w-[3.5em] mt-2 mx-auto">
+  <div v-if="!isLoginRoute && userStore.authenticated === undefined" class="w-[3.5em] mt-2 mx-auto">
     <Icon
       name="ei:spinner-3"
       size="3.5em"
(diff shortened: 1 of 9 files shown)


client/app.vue
```diff
@@ -1,98 +1,24 @@
 <template>
```
holloway, line 1: Previously a lot of template stuff was in the `app.vue` but I've moved it to a `layouts/default.vue` so that we can optionally hide it.

client/stores/user.ts
```diff
@@ -1,7 +1,11 @@
 import { defineStore } from 'pinia'
+import { LOGIN_PATH, onLoginRoute } from '~/utilities/url'
 
 export type ProfileData = {
-  authenticated: boolean
+  /**
+   * Until we fetch their profile their authentication state is unknown which we'll express as `undefined`
+   */
+  authenticated: undefined | boolean
```
holloway, line 8: This is the main logic change. Until we fetch the profile we don't know whether we're logged in, so there are 3 possible auth states not 2.

jennifer-richards: Seems reasonable to me. I suppose we could squabble about whether undefined or null is the better third state. I'm personally inclined toward null due to a fear of the undefined, but don't feel strongly. I only mention it because we have a null value serving a similar purpose beside it.

I guess this is different in that it's "don't know yet" versus "no value" since id would be null if authenticated were false.

client/stores/user.ts
```diff
@@ -14,17 +18,25 @@ type PretendingToBe = {
...
+      authenticated: undefined,
       id: null,
-      name: 'Guest',
+      name: '',
       email: '',
       avatar: '',
       rpcPersonId: null,
       isManager: false,
-      pretendingToBe: null // demo/debug only!
+      pretendingToBe: null // demo/debug only! FIXME: disallow on prod?
```
jennifer-richards, line 39: We can probably do away with it sooner than that - it had a limited scope to start with and hasn't been maintained. People can just log in as other real users now so I think it's outlived its usefulness.

jennifer-richards (commented):
lgtm

PR #276, opened 2025-05-08
feat: enable django admin, add models to admin

rudimatz:
what was done:
- on login to purple, grant Admin permission if DT user's role tuple ["leadmaintainer", "tools"] is set

still to do:
- build static files for dev and prod (run `./manage.py collectstatic`) - needs currently to be done manually
- discuss proper permissions to a) access purple, b) access Django admin


jennifer-richards (approved):
lgtm

PR #282, opened 2025-05-09
feat: add Django admin search fields and list display, apply black

rudimatz:


.devcontainer/devcontainer.json
```diff
@@ -7,7 +7,8 @@
 	// The .devcontainer/docker-compose.yml file contains any overrides you need/want to make.
 	"dockerComposeFile": [
 		"../docker-compose.yml",
-		"docker-compose.extend.yml"
+		"docker-compose.extend.yml",
+    "docker-compose.extend-custom.yml"
```
jennifer-richards, line 11: indentation lint 🧹

jennifer-richards (changes requested):
Looks good. One very minor lint to fix.

We'll probably need to add `raw_id_fields` to models that have FKs to large data tables (`Document`, `RfcToBe`, ...). By default, the admin will fetch all those options and give you a less-than-convenient drop-down select widget to choose the one you want.

That could and perhaps should be a separate PR, though.

jennifer-richards (approved):

PR #283, opened 2025-05-09
feat: add is_active flag for RpcPerson

rudimatz:
- added flag `is_active` for RfcPersons to be able to hide specific Persons from UI without deleting them

- modified existing endpoint:
~~GET `/api/rpc/rpc_person/` returns only active persons~~
~~GET `/api/rpc/rpc_person/?all=true` returns ALL persons~~

GET `/api/rpc/rpc_person/` returns all active persons
GET `/api/rpc/rpc_person/?is_active=true` returns ONLY active persons

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -114,12 +114,6 @@ class RpcPersonViewSet(viewsets.ReadOnlyModelViewSet, viewsets.GenericViewSet):
     filter_backends = (filters.DjangoFilterBackend,)
     filterset_fields = ["is_active"]
 
-    def get_queryset(self):
-        queryset = RpcPerson.objects.all()
-        if self.request.query_params.get("all") in ("true", "1"):
-            queryset = queryset.filter(is_active=True)
-        return queryset
-
     @with_rpcapi
     def get_serializer_context(self, rpcapi: rpcapi_client.DefaultApi):
         """Add context to the serializer"""


jennifer-richards (changes requested):
I think this'll work. However, I don't think it will document the "all" option in the OpenAPI spec. Can you look into doing that? It may require custom annotation (which is fairly common).

rjsparks:
Would it be hard to avoid using a queryset argument for this? Should the preference not be in the body of the request? Maybe what we really want is two endpoints?

jennifer-richards:
> Would it be hard to avoid using a queryset argument for this? Should the preference not be in the body of the request? Maybe what we really want is two endpoints?

It could be two endpoints, a POST, or whatever we want. My inclination in this case was that it's a simple enough parameter that it's ok this way.

Probably the best alternative would be a separate viewset for the unfiltered query set. That would require less custom code for drf.

rudimatz:
update:
refactor for more intuitive solution using Django filters, default keeps the current behavior, query param `?is_active=true` allows to request only active persons

rpc/api.py
```diff
@@ -110,6 +111,14 @@ def profile_as_person(request, rpc_person_id):
 class RpcPersonViewSet(viewsets.ReadOnlyModelViewSet, viewsets.GenericViewSet):
     serializer_class = RpcPersonSerializer
     queryset = RpcPerson.objects.all()
+    filter_backends = (filters.DjangoFilterBackend,)
+    filterset_fields = ["is_active"]
+
+    def get_queryset(self):
```
jennifer-richards, line 117: I think this method goes away since you've added the filter?

jennifer-richards (changes requested):
Need to remove the `get_queryset()` method, I think, otherwise looks good

rpc/api.py
rudimatz: removed it

jennifer-richards (approved):

PR #294, opened 2025-05-14
chore: ruff formatting and adding ruff to workflow tests

rudimatz:
- setting minimal configuration for ruff (line-length = 88)
- reformat code to make it ruff compliant
- add ruff format validation in workflow tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [tool.ruff]
 line-length = 88
-
+target-version = "py312"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[tool.ruff]
+line-length = 88
+
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,5 +20,5 @@ jobs:
 
     - name: Check code style
       run: |
-        pip install black
-        black --check --diff .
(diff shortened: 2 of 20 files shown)


pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[tool.ruff]
+line-length = 88
```
kesara, line 2: `www` has following for ruff config: https://github.com/ietf-tools/www/blob/de97a842ffff660c18eaa10f22824914dcd2641d/pyproject.toml#L11-L26

Is there any thing we can import from that?

jennifer-richards: Looks to  me like those are configs for the linter and hints to save time looking at python files we don't need to format. Since this PR only uses `ruff format` they won't affect it. If/when we turn on more of ruff's checks, we should remember to look back at that.

jennifer-richards (approved):

pyproject.toml
jennifer-richards: (well, to be precise, the hints about which files to look at probably would affect the format check, but I don't think there's any need to be selective in this project at this point)

rudimatz: @kesara I added the target-version to the config in latest commit, otherwise - as Jennifer said - lets keep it lean and add lint-configs once we need it

Issue #91, opened 2024-01-03
"My Documents" data fetch needs overhaul


jennifer-richards:
The "My Documents" page loads data to display by retrieving _all_ the `RfcToBe` data and _all_ assignments from the backend, then filters these on the client side. This is extremely slow (takes ~11 seconds on my dev machine with the migrated data set) and reveals data that should perhaps not be visible to every client.

A substantial chunk of this time (around 75-80%) is spent computing history that is not actually shown on the view. (This is a bit worrying because there's essentially no history in the migrated data, but it's not time to panic about it.)

microamp:
I can reproduce the behaviour locally. When I refresh the "My Documents" pages. The client makes the following API calls.

- <http://localhost:8088/api/rpc/rpc_person/>
- <http://localhost:8088/api/rpc/assignments/>
- <http://localhost:8088/api/rpc/labels/>
- <http://localhost:8088/api/rpc/profile/>
- <http://localhost:8088/api/rpc/documents/>

All calls took less than a second except the documents one which took 15+ seconds.

I also have gathered database logs for the page load. It's in a file called `db.log` below.

``` bash
head -5 db.log
```

``` example
2024-09-12 22:47:36.258 UTC [156] LOG:  statement: SELECT set_config('TimeZone', 'UTC', false)
2024-09-12 22:47:36.260 UTC [156] LOG:  statement: SELECT "django_session"."session_key", "django_session"."session_data", "django_session"."expire_date" FROM "django_session" WHERE ("django_session"."expire_date" > '2024-09-12T22:47:36.254995+00:00'::timestamptz AND "django_session"."session_key" = 'hl4gjb6qbn3enolzowalxyixce22h8bc') LIMIT 21
2024-09-12 22:47:36.263 UTC [156] LOG:  statement: SELECT "rpcauth_user"."id", "rpcauth_user"."password", "rpcauth_user"."last_login", "rpcauth_user"."is_superuser", "rpcauth_user"."username", "rpcauth_user"."first_name", "rpcauth_user"."last_name", "rpcauth_user"."email", "rpcauth_user"."is_staff", "rpcauth_user"."is_active", "rpcauth_user"."date_joined", "rpcauth_user"."name", "rpcauth_user"."datatracker_subject_id", "rpcauth_user"."avatar" FROM "rpcauth_user" WHERE "rpcauth_user"."id" = 1 LIMIT 21
[...]

PR #297, opened 2025-05-15
refactor: improve fetching my documents

rudimatz:
Backend: 
new behavior: 
- only assignments of currently authenticated user are returned for endpoints `GET /api/rpc/assignments` 
- assignments in payload also contain embedded data of related RfcToBe

Frontend:
Improve API calls on page "My Documents (Documents assigned to me)", requesting all documents and all assignments and intersecting is no longer required

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -50,6 +50,7 @@ from .serializers import (
     StreamNameSerializer,
     TlpBoilerplateChoiceNameSerializer,
     VersionInfoSerializer,
+    UserSerializer,
 )
 from .utils import VersionInfo
 
@@ -263,6 +264,20 @@ class ClusterViewSet(viewsets.ReadOnlyModelViewSet):
 class AssignmentViewSet(viewsets.ModelViewSet):
     queryset = Assignment.objects.all()
     serializer_class = AssignmentSerializer
+
+    def get_queryset(self):
+
+        user = self.request.user
+
+        person_id = UserSerializer().get_person_id(user)
+
+        if person_id is None:
+            return Assignment.objects.none()
+
+        # Filter assignments for the logged-in RpcPerson
+        return Assignment.objects.filter(person_id=person_id).select_related(
+            "rfc_to_be"
+        )
 
 
 class RfcToBeViewSet(viewsets.ModelViewSet):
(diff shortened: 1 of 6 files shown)


rudimatz:
fixes https://github.com/ietf-tools/purple/issues/91

rpc/api.py
```diff
@@ -264,6 +266,24 @@ class AssignmentViewSet(viewsets.ModelViewSet):
     queryset = Assignment.objects.all()
     serializer_class = AssignmentSerializer
 
+    def get_queryset(self):
+        user = self.request.user
+
+        base_queryset = Assignment.objects.select_related("rfc_to_be")
+
+        if user.is_superuser:
```
jennifer-richards, line 274: Could you also grant full access to people with the "manager" RpcRole?

https://github.com/ietf-tools/purple/blob/e41d6de149fb9aa6ede827b96ba9780ec640db9f/rpc/migrations/0003_populate_rpcrole.py#L24

rudimatz: added in latest commit

rpc/serializers.py
```diff
@@ -56,6 +56,14 @@ def get_person_id(self, user) -> Optional[RpcPerson]:
             return rpc_person.pk
         return None
 
+    def has_role(self, user, role) -> Optional[RpcPerson]:
```
jennifer-richards, line 59: The type hints are not correct, the return value is a boolean.

This should be factored out of the serializer class and into a helper method. It'd be reasonable to put it the root of this class for now. It'll probably end up in a module with other authz-related methods but it's ok in this module until we know what else will be there.

Python thing: if you want to keep it in this class but use it in other places as well, it should be a static method. Since it doesn't rely on `self`, you can just remove that argument and add the `@static` decorator to the method. Then, other places can call it on the class like `UserSerializer.has_role(...)` without instantiating the serializer. That's preferable to making a throwaway instance. In this case, though, I don't think the `has_role()` check belongs to the serializer.

jennifer-richards (changes requested):
One bit of refactoring suggested inline.

rpc/serializers.py
rudimatz: @jennifer-richards ok new commit pushed

jennifer-richards (approved):

PR #305, opened 2025-05-20
feat: refactor how approvals are captured

rjsparks:


rpc/models.py
```diff
@@ -313,19 +312,77 @@ def __str__(self):
...
+                    models.Q(overriding_approver__isnull=True)
+                    | models.Q(approver__isnull=False)
+                ),
+                name="finalapproval_approval_override_requires_approver",
+                violation_error_message="approval override requires an approver be set",
+            ),
+            models.CheckConstraint(
+                check=(models.Q(body="") | models.Q(overriding__approver__isnull=True)),
+                name="finalapproval_body_approval_no_override",
+                violation_error_message="body approval cant be overridedn",
```
jennifer-richards, line 383: Typo

```suggestion
                violation_error_message="body approval cant be overridden",
```

rpc/models.py
```diff
@@ -313,19 +312,77 @@ def __str__(self):
...
+    be not None. Overriding approver will be None when body is used.
+
+    Overriding approver is used when someone has to step in for whoever is
+    pointed to as approver (such as when an author passes away or otherwise
+    becomes non-responsive). Overriding approver should never be not None if
+    approver is None.
+
+    Some approvals are Auth48 approvals from authors. Any FinalApproval with
+    an approval pointing to the same DatatrackerPerson as a RfcAuthor for that
+    RfcToBe is an Auth48 approval.
```
jennifer-richards, line 334: I think "and an empty body" is needed to declare something an Auth48 approval to cover the (unlikely and perhaps procedurally dubious but not technically impossible) case that someone who is an author also approves a body approval

jennifer-richards (changes requested):

rpc/models.py
rjsparks: the key there is procedurally dubious - it won't happen because leadership will watch for that kind of conflict and not let it happen. I'm willing to add "and and empty body" to the docstring, but I dont think it's worth trying to code that into the constraints.

jennifer-richards: (capturing comment from DM for visibility/posterity) That's all I expected. If we later add a manager to provide `FinalApproval.objects.auth48()`, it should perhaps filter out ones with a body.

jennifer-richards (approved):

PR #313, opened 2025-05-27
feat: ApprovalLogMessage and Label.is_complexity

rjsparks:


rpc/models.py
```diff
@@ -314,24 +314,20 @@ def __str__(self):
...
 
-    The RPC may need to ask a body, like the IESG, for approval without knowing
-    which person will provide that approval. Body would be non-blank, approver
-    would be None, approved would be None. The body will have a person that
-    approves on their behalf, so once approved is not None, approver will also
-    be not None. Overriding approver will be None when body is used.
+    Sometimes the titlepage author is not a person, such as when it is the IAB.
+    The request for approval from such a body is captured in the body field.
+    Body would be non-blank, approver would be None, approved would be None.
+    A person will provide approval on belahf of that body, so once approved
```
jennifer-richards, line 323:
```suggestion
    A person will provide approval on behalf of that body, so once approved
```

rpc/models.py
```diff
@@ -584,3 +581,28 @@ def __str__(self):
...
+
+class ApprovalLogMessage(models.Model):
+    """Public log of final approval-related steps
+
+    These messages will be displayed on the approvals
+    views (historically the AUTH48 approvals page) and
+    will be publically visible.
+    """
+
+    rfc_to_be = models.ForeignKey(RfcToBe, on_delete=models.PROTECT)
```
jennifer-richards, line 594: Intentional to let an ApprovalLogMessage block deletion of an RfcToBe? I'd have expected it to be the other way around with other safeguards preventing such deletion, which should be extremely rare.

(This is not an objection, just checking - it should be rare enough that maybe the admin having to explicitly remove the log messages before doing the delete is useful as a safeguard)

jennifer-richards (changes requested):
question and a typo

rpc/models.py
rjsparks: I didn't put a lot of thinking into this - I just took RpcComment's approach.

PR #314, opened 2025-05-27
fix create assignment

rudimatz:


rudimatz:
@jennifer-richards refactored code to
- revert back Assignments serializer to default behavior
- revert client to default params
- inject nested RfcToBe data only in RpcPersonAssignmentViewSet for GET

PR #322, opened 2025-05-29
feat: memcache+caching for DatatrackerPerson.plain_name()

jennifer-richards:


rjsparks (approved):

rjsparks (approved):

jennifer-richards (commented):
Oops, I need to make a couple more changes:

1. Specify a lifetime for the cache entry 
2. Fill in the rest of the production caches spec

jennifer-richards:
Ok, I think I'm done. Rather than putting an explicit timeout on the cache call, I set the default timeout to 10 minutes. It had been 5 minutes.

I also disabled caching in dev by commenting out the config block. I think for most dev / debugging, we'll want the cache out of the way.

Re-requested review just to confirm we're comfortable with those choices.

PR #328, opened 2025-05-30
chore: fix ietf copyright comments

rudimatz:


rpc/migrations/0003_populate_rpcrole.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 4.2.4 on 2023-08-24 17:31
+# Copyright The IETF Trust 2023, All Rights Reserved
```
rjsparks, line 1: make these 2023-2025

rjsparks (approved):

PR #337, opened 2025-06-06
feat: document comments API + history

jennifer-richards:
This adds APIs for listing, creating, and updating comments on an RfcToBe or Document (i.e., datatracker draft that does not yet have an RfcToBe). List and create are done via the `documentsCommentList()` and `documentsCommentsCreate()` API calls in the generated TypeScript API client. Updates can be done with `documentsCommentsUpdate()` and `documentsCommentsPartialUpdate()` (though both accept only the "comment" field, so they're interchangeable). 

image

Creating a comment is done by POSTing a comment. All fields except the comment text are set automatically to the current user / time.

Listing comments returns the comment id, comment text, timestamp, the `plainName` and `RpcPerson` PK of the comment author (in the `by` field). If the comment has been edited, the `last_edit` field indicates the time of the edit and the name of the editor, otherwise `last_edit` is null. Results are in reverse time order and can be paginated using `?limit=x&offset=y`. The response JSON looks like this:
```json
[
  {
    "id": 4,
    "comment": "modify!?",
    "by": {
      "name": "Jennifer Richards",
      "rpcperson": 12,
      "avatar": null
    },
    "time": "2025-06-09T17:57:33.981007Z",
    "last_edit": {
      "by": {
        "name": "Jennifer Richards",
        "person_id": 13,
        "avatar": "https://www.ietf.org/[...]"
      },
      "date": "2025-06-09T18:11:31.054105Z"
    }
  },
  {
    "id": 3,
    "comment": "Post-pub comment",
    "by": {
      "name": "Jennifer Richards",
      "rpcperson": 12,
      "avatar": null
    },
    "time": "2025-06-09T15:05:24.975565Z",
    "last_edit": null
  },
  {
    "id": 2,
    "comment": "Pre-pub comment",
    "by": {
      "name": "Jennifer Richards",
      "rpcperson": 12,
      "avatar": null
    },
    "time": "2025-06-09T15:04:51.697814Z",
    "last_edit": null
  }
]
```

The current modeling ties a comment to a DatatrackerPerson - we may change this, but I have not done so yet. For now, the `rpcperson` field may be null, as shown in the second comment above. This _probably_ won't happen for comments created through the API because only users with an RpcPerson are allowed to comment. It's possible to create them programmatically or to delete an RpcPerson, though, so front end code should allow for the possibility that it'll get a null. (The reason for including the rpcperson ID is to allow linking to the Team page of the commenter, if desired)

Permissions are not implemented yet, beyond checking that the user is logged in and has an RpcPerson.


rudimatz (approved):

jennifer-richards:
> The clientside's template has space for a profile image. I see the `avatar` URL in the profile but do we have that in the comment api?

It's not there now but could be added. We won't have an avatar for all commenters, since backfilled data will include people who won't have a user account, so it'll sometimes be null.

rjsparks (approved):

rjsparks (approved):

PR #343, opened 2025-06-11
feat: add constraint for rfc number unique, enable history display in…

rudimatz:
… django admin


rpc/models.py
```diff
@@ -112,6 +112,15 @@ def __str__(self):
             f"RfcToBe for {self.draft if self.rfc_number is None else self.rfc_number}"
         )
 
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=['rfc_number'],
+                name='unique_non_null_rfc_number',
+                condition=models.Q(rfc_number__isnull=False)
```
jennifer-richards, line 120: Don't need a `condition`. Postgres (and mysql and most other databases) ignore nulls when enforcing unique constraints.

With Django 5.0+ (true for purple, not yet for datatracker) you can use the `nulls_distinct` parameter if you want to be explicit (good idea) rather than relying on the database default. https://docs.djangoproject.com/en/5.2/ref/models/constraints/#nulls-distinct

rudimatz: added explicit definition `nulls_distinct=True` (available in Postgres 15+)

PR #356, opened 2025-06-19
fix: blank=True for several RpcPerson fields

jennifer-richards:
Allows RpcPerson to be created/managed in the admin. Does not affect the database schema.

### DIFF
--- a/rpc/models.py
+++ b/rpc/models.py
@@ -30,11 +30,12 @@ class RpcPerson(models.Model):
     datatracker_person = models.OneToOneField(
         "datatracker.DatatrackerPerson", on_delete=models.PROTECT
     )
-    can_hold_role = models.ManyToManyField("RpcRole")
-    capable_of = models.ManyToManyField("Capability")
+    can_hold_role = models.ManyToManyField("RpcRole", blank=True)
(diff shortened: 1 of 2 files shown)


rudimatz (approved):

rpc/models.py
```diff
@@ -30,11 +30,12 @@ class RpcPerson(models.Model):
...
         "datatracker.DatatrackerPerson", on_delete=models.PROTECT
     )
-    can_hold_role = models.ManyToManyField("RpcRole")
-    capable_of = models.ManyToManyField("Capability")
+    can_hold_role = models.ManyToManyField("RpcRole", blank=True)
+    capable_of = models.ManyToManyField("Capability", blank=True)
     hours_per_week = models.PositiveSmallIntegerField(default=40)
     manager = models.ForeignKey(
         "RpcPerson",
+        blank=True,
```
rudimatz, line 38: isn't null-True sufficient here (if there is no manager)? when would it be blank?

jennifer-richards: Basic rule for most situations in Django is: for anything but `CharField` / `TextField`, you want `blank` and `null` to be equal. For `CharField` or `TextField`, you always want `null=False` so that `blank` is the empty string.

PR #358, opened 2025-06-20
feat: store references on import

rudimatz:
for new drafts imported into purple (endpoint `POST /api/rpc/submissions/{draft_id}/import/`), add logic:

- check if document {draft_id} has normative references returned from DT-Api `get_draft_references`
- check if RfcToBe for those referenced docs already exist in purple
- if yes: ignore them
- if no: it's a missref, create the doc (if it doesn't exist yet) and create new RpcRelatedDocument of relationship "missref" for the doc

also:
- add migration to populate DocRelationshipName  for `missref` 
- add Viewset for RpcRelatedDocument for endpoint `GET /api/rpc/documents/{draft_name}/references/`, returns e.g.
```[
    {
        "relationship": "missref",
        "source": 47,
        "target_document": 37,
        "target_rfctobe": null
    },
    {
        "relationship": "missref",
        "source": 47,
        "target_document": 38,
        "target_rfctobe": null
    }
]

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -305,7 +305,7 @@ def import_submission(request, document_id, rpcapi: rpcapi_client.DefaultApi):
                     except Document.DoesNotExist:
                         draft_info = rpcapi.get_draft_by_id(reference.id)
                         if draft_info is None:
-                            return Response(status=404)
+                            raise NotFound("Unable to get draft info for reference")
                         draft, _ = Document.objects.get_or_create(
                             datatracker_id=reference.id,
                             defaults={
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: rpc/api.py [hunk 1]
```diff
@@ -274,9 +276,51 @@ def import_submission(request, document_id, rpcapi: rpcapi_client.DefaultApi):
         )
 
     # Create the RfcToBe
-    serializer = CreateRfcToBeSerializer(data=request.data, context={"draft": draft})
+    serializer = CreateRfcToBeSerializer(data={**request.data, "draft": draft.pk})
```
jennifer-richards, line 279: An alternative syntax for dict merging is
```python
data=request.data | {"draft": draft.pk}
```
(I find it somewhat more readable, though not everyone is familiar with it)

jennifer-richards (approved):
lgtm, thanks for the adjustments

(I think there are probably api client generation warnings but we can clean those up separately.)

PR #364, opened 2025-06-25
fix: revert relation rfctobe to doc foreignkey

rudimatz:
should be - for now - kept as ForeignKey

OneToOne-relation requires further work on xfer script (using history)


jennifer-richards (approved):
Keep in mind that existing instances will need to be torn down since this removes rather than reverses the migration.

PR #366, opened 2025-06-26
feat: Add ability to send emails

kesara:


jennifer-richards (changes requested):
Looks good. Before merge, though, it'd be nice to put in a guard against trying to send a message when we haven't explicitly configured the backend. By default it'll try to connect to localhost:25, which fails / raises a connection error. That'll usually fail, but we could give clearer feedback as to why, and it leaves the chance that we somehow emit messages we didn't intend to emit.

rjsparks:
Would this be a startup check? Just to see if the env variables exist?

purple/mail.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def send_mail(to, subject, msg, frm=None, fail_silently=True):
+    if not frm:
+        frm = settings.DEFAULT_FROM_EMAIL
+    if isinstance(to, str):
+        to = [
+            to,
+        ]
+    dj_send(subject, msg, frm, to, fail_silently=fail_silently)
```
kesara, line 14: I tried to write test for this but I couldn't find a sane way to mock this properly to test.

jennifer-richards (approved):

PR #368, opened 2025-06-26
fix: navigate to enqueuing page after initial ingest

rjsparks:
Do we have any way to reverse paths so we're not hardcoding them into .vue files?

This stays with the current convention of not using trailing slashes.


holloway:
In Red we have a [`url.ts`](https://github.com/ietf-tools/red/blob/main/client/utilities/url.ts) with constants and url path builder functions etc. Perhaps we could do that too?

PR #376, opened 2025-07-02
feat: add order endpoint for RfcAuthors

rudimatz:
- Add migration for field "order" on RfcAuthors, default: 0
- Add default ordering of Authors in AuthorsList by new field "order"

- Create endpoint POST `/api/rpc/documents/{draft_id}/authors/order/`
that accepts payload like 
```{
    "order": [
        47897,
        23950
    ]
}
```
First item in array => First author of doc
...

Before updating the order, the function checks that authors currently set and those passed in Array are identical (besides the order)

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -463,7 +463,7 @@ class RpcAuthorViewSet(viewsets.ModelViewSet):
         serializer.save(rfc_to_be=rfc_to_be, order=max_order + 1)
 
     def get_serializer_class(self):
-        if self.action in ("create"):
+        if self.action == "create":
             return CreateRfcAuthorSerializer
         return RfcAuthorSerializer
 


jennifer-richards (changes requested):
See inline - I think this needs a transaction and a bit of refactoring.

rpc/api.py
```diff
@@ -454,7 +454,14 @@ def perform_create(self, serializer):
         ).first()
         if rfc_to_be is None:
             raise NotFound("RfcToBe with the given draft name does not exist")
-        serializer.save(rfc_to_be=rfc_to_be)
+        # Find the current highest order for this document
+        max_order = (
+            RfcAuthor.objects.filter(rfc_to_be=rfc_to_be)
+            .aggregate(max_order=Max("order"))
+            .get("max_order")
+            or 0
```
jennifer-richards, line 462: trivial: you can pass `default=0` to `Max()` and avoid the `or 0`

jennifer-richards (commented):
One trivial change I'd suggest, though I won't insist.

Did you check that the openapi schema is correct for the update / partialUpdate operations? I suspect that the `get_fields()` will fly under the radar.

jennifer-richards (approved):

PR #389, opened 2025-07-10
feat: change reference creation based on DRF changes

rudimatz:
- move `create_draft_by_name` logic to utils to make it reusable
- add check if doc already exists in purple, if not, call DT to get doc infos and create it


rpc/serializers.py
```diff
@@ -350,13 +351,43 @@ def get_target_draft_name(self, obj: RpcRelatedDocument) -> str:
...
+            target_rfctobe = RfcToBe.objects.filter(
+                draft__name=target_draft_name
+            ).first()
+            if not target_rfctobe:
+                raise serializers.ValidationError(
+                    f"No Document or RfcToBe found for draft name '{target_draft_name}'"
+                )
+
+        try:
+            related_doc = RpcRelatedDocument.objects.create(
```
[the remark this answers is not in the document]
rudimatz: yes, but it would only return payload like
```{
    "id": 56,
    "relationship": "missref",
    "source": 1
}```

without the "target_draft_name".
I will push new solution that will fix that

File: rpc/serializers.py [hunk 2]
```diff
@@ -350,13 +351,64 @@ def get_target_draft_name(self, obj: RpcRelatedDocument) -> str:
...
+        ret = super().to_representation(instance)
+        ret["target_draft_name"] = ret.pop("target_draft_name_output", None)
+        return ret
+
+    def create(self, validated_data):
+        target_draft_name = validated_data.pop("target_draft_name")
+
+        source = validated_data["source"]
+
+        target_document = Document.objects.filter(name=target_draft_name).first()
```
jennifer-richards, line 387: You need to check for an RfcToBe first. If there is an RfcToBe, there will always be a Document, so this will always link to a Document or fail validation.

rpc/utils.py
```diff
@@ -32,21 +34,41 @@ def next_rfc_number(count=1) -> list[int]:
...
-    else:
-        raise ValueError(f"Invalid target type: {target_type}")
 
     serializer = CreateRpcRelatedDocumentSerializer(data=data)
     if serializer.is_valid(raise_exception=True):
         return serializer.save()
     return None
+
+
+def get_or_create_draft_by_name(draft_name, *, rpcapi):
```
jennifer-richards, line 50: Can this also be used to replace the similar code in `import_submission` rather than maintaining it twice? (i.e., the call on line 318 of rpc/api.py)

jennifer-richards (changes requested):
I think there's one must-fix issue, plus a small additional refactor you might (or might not) make. Apologies for not catching these on the first pass.

The serializer wrangling looks like it'll work, but should at least have a comment explaining what it's doing with the _output field and to_representation hacking.

(I'm also wondering whether it'd be better to go back to separate request/response serializers, I hadn't considered that it'd require this much cleverness to make it work.)

rpc/utils.py
rudimatz: as discussed, I'll leave as is because of different inputs (draft-name vs draft-id)

PR #394, opened 2025-07-11
fix: on put and patch for author, update only validated fields, not order

rudimatz:
try to avoid a full update on all fields of author on patch/put, but only update the fields that were send


jennifer-richards (approved):
Nice. We'll have to remember to adjust this if we start adding m2m fields to the model, but that's not likely to come up.

PR #405, opened 2025-07-18
fix: use Django's logout page for sign out

jennifer-richards:
Navigates to `/login/` instead of using a JS fetch to sign out. The `/login/` page currently handles both login/logout.

WIP: the `/login/` path should be renamed, and the page is ancient proof-of-concept. This is a proof of concept right now, not at all ready for merge.


rudimatz (approved):

jennifer-richards:
OBE (#577)

PR #407, opened 2025-08-01
feat: DocRelationshipNames viewset

jennifer-richards:
Adds `DocRelationshipNamesList` and `DocRelationshipNamesRetrieve` API calls. Small refactor to use a common `NameSerializer` for subclasses of `models.Name` instead of maintaining a fleet of serializers identical but for name and model name.


holloway (approved):
lgtm

PR #408, opened 2025-08-04
feat: Document dependency relationship list

holloway:
## feat

* Building on #407 this adds a UI for selecting those names in a dropdown

Should the table of references have its values formatted in a different way? The values look wrong to me :shrug: 

Screenshot_2025-08-04_15-42-16
Screenshot_2025-08-04_15-40-46
Screenshot_2025-08-04_15-40-50


jennifer-richards:
> Should the table of references have its values formatted in a different way? The values look wrong to me 🤷

Yes, the values in those columns don't make any sense relative to the header. The "not-a-design" wireframe for that looks like

image

I'll create a new issue to track cleaning that up

PR #411, opened 2025-08-06
feat: Aggregate queue tabs and add filter

holloway:
## feat

* Merge queue tabs 'pending assignment', 'exceptions', and 'in process' into a new tab 'queue'
* add filters (I don't know how to move documents to this tab so I haven't tested this)
* vertically align `DocumentTable` `<th>`s

**screenshot**
Screenshot_2025-08-06_14-43-47

### DIFF
--- a/client/pages/queue/[section].vue
+++ b/client/pages/queue/[section].vue
@@ -37,6 +37,26 @@
       </button>
     </div>
 
+    <div v-if="currentTab=== 'queue'">
+      <fieldset>
+        <legend class="font-bold">Filters</legend>
+        <RpcCheckbox
+          id="needsAssignment"
+          label="Needs Assignment?"
+          size="small"
+          :checked="needsAssignment"
+          @change="(e) => needsAssignment = extractChecked(e)"
+        />
+        <RpcCheckbox
+          id="hasException"
+          label="Has Exception?"
+          size="small"
+          :checked="hasException"
+          @change="(e) => hasException = extractChecked(e)"
+        />
+      </fieldset>
+    </div>
+
     <!-- DATA TABLE -->
 
     <div class="mt-2 flow-root">
@@ -86,7 +106,7 @@ const api = useApi()
 
 // DATA
 
-const tabs: Tab[] = [
+const tabs = [
   {
     id: 'submissions',
     name: 'Submissions',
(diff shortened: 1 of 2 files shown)


client/pages/queue/[section].vue
```diff
@@ -344,31 +357,29 @@ const filteredDocuments = computed(() => {
...
+        .filter(
+          (d: any) => {
+            if(needsAssignment.value) {
+              return d.assignmentSet?.length === 0
+            }
+            if(hasException.value) {
+              return !!d.exception
+            }
+            return true
+        })
```
Comment (rjsparks) [inline client/pages/queue/[section].vue:375]: I'm not sure this is quite what we need yet. With this implementation, if I read it correctly, the needsAssignment filter will hide the hasException filter when selected. The first return wins and the second condition is never evaluated.


The assigner should be able to look for documents needing assignment, seeing only those that have exceptions (to assign to people who can handle them), only those that don't have exceptions (for simpler assignments), or the full set of documents needing assignment whether the documents have exceptions or not.

Similarly, all of the users should be able to look at the documents that have exceptions, or don't have exceptions, whether the document is assigned or not.

That is, each of these filters should have [yes, no, either] semantics.

rjsparks (changes requested):
This is definitely on the right track, but the filtering semantics need to be a little richer.

client/pages/queue/[section].vue
Comment (holloway) [reply client/pages/queue/[section].vue]: Ok I've added a 'tristate' checkbox (see video on PR) and updated the filters.

Is it possible to move documents to this tab yet?

Comment (rjsparks) [reply client/pages/queue/[section].vue]: documents should _definitely_ already be appearing in this tab.

Comment (jennifer-richards) [reply client/pages/queue/[section].vue]: I think the issue was a bug you fixed in your last commit (bca2068) and the absurd slowness of the `queueList()` API call. The latter is fixed on main and on xfer. Do you now see documents on the queue?

client/pages/queue/[section].vue
```diff
@@ -344,31 +361,43 @@ const filteredDocuments = computed(() => {
...
+              } else if(needsAssignmentTristate.value === false) {
+                return Boolean(d.assignmentSet?.length > 0)
+              } else if(needsAssignmentTristate.value === CHECKBOX_INDETERMINATE) {
+                return true
+              }
+            }
+
+            const hasExceptionFilterFn = () => {
+              if(hasExceptionTristate.value === true) {
+                return Boolean(d.exception)
```
Comment (jennifer-richards) [inline client/pages/queue/[section].vue:384]: Is `exception` getting set somewhere? It should be a check for an attached label with `is_exception` set. The back end doesn't currently do this. Should it?

I lean toward yes, probably as a `hasException` field, but that raises the question of whether it should also have an `exceptions` list. I'd rather not because it would (I guess) be redundant with the `labels` list... any thoughts?

client/pages/queue/[section].vue
Comment (holloway) [reply client/pages/queue/[section].vue]: @jennifer-richards yes, now I am

client/pages/queue/[section].vue
Comment (holloway) [reply client/pages/queue/[section].vue]: @jennifer-richards I've updated the filter again. Is that what it should look like? We can do the server approach instead if you want.

Comment (jennifer-richards) [reply client/pages/queue/[section].vue]: lgtm, and it's simple enough I see no need to complicate it futher :-)

jennifer-richards (approved):
Looks good to me. UX-wise I find the tristates a little peculiar but learnable and I don't have a clearly better suggestion. I suppose it could be sets of radio buttons with "yes," "no," and "don't care" labels, but that is less compact.

rjsparks:
Lets merge this and get it onto staging and get some feedback

PR #424, opened 2025-08-13
fix: capture and flag withdrawn references

rjsparks:


rjsparks:
Based on offline discussion, this isn't the right path. We'll instead create a new relationship name for withdrawn references.

PR #441, opened 2025-08-22
feat: label management improvements

jennifer-richards:
Feature-wise, this exposes the "used" flag through the "manage labels" page. Cleans up some reactivity / state management / error displays to be more in line with how we've evolved. There's more cleanup to be done.

Adds a very simple red "X" mark on the label list display to indicate labels with used = False. This will need to be improved, but it's more than I have time for now.

### DIFF
--- a/client/components/RpcLabelEditDialog.vue
+++ b/client/components/RpcLabelEditDialog.vue
@@ -102,11 +102,10 @@ type Props = {
 }
 
 const props = defineProps<Props>()
-const label = reactive(props.label)
+const label = reactive<Label>({ ...props.label })
 
-async function save () {
+async function save() {
   const labelData: Label = {
-    id: 0, // FIXME: is this ok?
     slug: label.slug,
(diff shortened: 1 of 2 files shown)


client/components/RpcLabelEditDialog.vue
```diff
@@ -96,27 +108,26 @@ if (!_overlayModalMethods) {
...
+  slug: '',
+  isException: false,
+  color: 'slate',
+  used: true
 }
 
-const props = defineProps<Props>()
-const label = reactive(props.label)
+const label = reactive<Label>(
+  props.label ? { ...props.label } : NEW_LABEL_DEFAULTS
```
holloway, line 122:
```suggestion
  { ...NEW_LABEL_DEFAULTS, ...props.label }
```

holloway (approved):
LGTM only one tiny suggestion that would make the defaults more granular if props.label was missing fields

PR #446, opened 2025-08-27
chore: allow multiple selections within label group

holloway:
## chore

*  allow multiple selections within label group (remove reka listbox, replace with rpccheckbox)


rjsparks:
nit - this is really a behavior change, so I think it should be squashed as a "fix: ".

PR #448, opened 2025-08-27
fix: remove duplicate columns and simplify logic

rudimatz:
remove duplicate column deadline
image

This commit removes the 2nd one


holloway (approved):
LGTM :+1:

PR #452, opened 2025-08-28
feat: show RPC member name, email and assignments

rudimatz:


client/pages/team/[id].vue
```diff
@@ -186,12 +130,98 @@
...
+const api = useApi()
+
+// Fetch person data
+const { data: person, pending, error } = await useAsyncData(
+  `person-${route.params.id}`,
+  async () => {
+    try {
+      const response = await api.rpcPersonRetrieve({ id: Number(route.params.id) })
+      return response
+    } catch (err) {
```
Comment (holloway) [inline client/pages/team/[id].vue:196]: I wonder if we should try/catch or just throw so that the `error` is set.

client/pages/team/[id].vue
```diff
@@ -72,37 +75,25 @@
...
-                </dd>
-              </div>
-              <div class="mt-4 flex w-full flex-none gap-x-4 px-6">
-                <dt class="flex-none">
-                  <span class="sr-only">Status</span>
-                  <Icon name="uil:times" class="h-6 w-5 text-gray-400" aria-hidden="true"/>
-                </dt>
-                <dd class="text-sm leading-6 text-gray-500">123</dd>
+              <div class="flex-none self-end px-6 pt-4">
+                <dt class="sr-only">Roles</dt>
```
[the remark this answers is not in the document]
Comment (rudimatz) [reply client/pages/team/[id].vue:93]: yeah I mostly copy and pasted. I would consider this just a mockup what we could display, not final layout.

client/pages/team/[id].vue
Comment (rudimatz) [reply client/pages/team/[id].vue:196]: @holloway just to understand, what is your concern with try/catch?

Comment (holloway) [reply client/pages/team/[id].vue:196]: @rudimatz that it will hide any errors so the return value of ` { data: person, pending, error }` will never have `error` set for us to use in the Vue template. It's ok to let the API client throw exceptions in a `useAsyncData` afaik.

Alternatively when you catch it you could pass the `err` to `snackbarForErrors()`

rjsparks:
yeah - I would not take the vertical space even unless there was something to show?

rudimatz:
@holloway see commit ~~ [typescript class RpcPerson instead any](https://github.com/ietf-tools/purple/pull/452/commits/aaf3d60261b695e87cedec2a44328bee0f91f319)

client/pages/team/[id].vue
Comment (rudimatz) [reply client/pages/team/[id].vue]: ok fixed it

holloway (approved):

PR #462, opened 2025-09-08
feat: add submitted and published in model and UI

rudimatz:
- populate "recently published" tab with RFCs published in last 30 days, starting with most recent
- show submitted date in queue page
- add `submitted_at` and `published_at` fields in model
- make errata type optional (allow null)

### DIFF
--- a/errata/migrations/0001_initial.py
+++ b/errata/migrations/0001_initial.py
@@ -124,9 +124,7 @@ class Migration(migrations.Migration):
                 (
                     "type",
                     models.ForeignKey(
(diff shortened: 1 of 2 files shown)


jennifer-richards (changes requested):
One more change - I didn't notice it before, but @rjsparks mentioned it in our chat earlier. The changes in the errata app should come out of this PR. (Nothing wrong with them, but unless I'm missing something they're a separate concern.)

rjsparks (approved):

PR #463, opened 2025-09-08
chore: nuxt links

holloway:
## chore

* replace all Vue `<a>` and `<NuxtLink>`s with an `<Anchor>` wrapper that can decide SPA navigation or not. This also ensures that external links have the `rel="noopener"` for security reasons. There's a `data-link-type` attribute rendered on each `<Anchor>` for debug, so you can use browser dev tools to check which link type is being used. We can remove that later as it doesn't do anything functional (it's only for debug).

### DIFF
--- a/client/components/Anchor.vue
+++ b/client/components/Anchor.vue
@@ -1,6 +1,6 @@
 <template>
   <NuxtLink
-    v-if="isInternal && !isHash && !isMailTo"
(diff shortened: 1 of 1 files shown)


jennifer-richards (approved):
Question inline, not a request for change

rjsparks (approved):

PR #464, opened 2025-09-09
chore(deps): bump 21 package versions

jennifer-richards:
This is probably mostly covered by a dependabot PR, but that also includes a nuxt version bump that's a bit more involved.


holloway (approved):
LGTM

holloway:
arguably we should move the `@types/*` packages to dev deps but that can happen anytime later

PR #478, opened 2025-09-11
feat: add comment widget

rudimatz:
this PR adds comment widget on the submission page of a draft, and on the detail page of a draft in the queue

it also fixes a bug: when you submitted a draft and reach the enqueue page of a draft, and then add a comment, the comments will disappear. after a reload, they reappear. The fix triggers the id-watcher, but the solution should be discussed.

### DIFF
--- a/client/app/pages/docs/[id]/index.vue
+++ b/client/app/pages/docs/[id]/index.vue
@@ -194,11 +194,12 @@ const {
   refresh: commentsReload
 } = await useCommentsForDraft(draftName.value)
 
-const { data: history, error: historyError, status: historyStatus, refresh: historyRefresh } = await useAsyncData(
-  () => `history-${draftName.value}`,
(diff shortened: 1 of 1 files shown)


jennifer-richards (approved):

client/app/pages/docs/import.vue
```diff
@@ -105,6 +105,25 @@
               </div>
             </div>
           </div>
+          <!-- Comments -->
+          <div v-if="!backendPending" class="mt-10 grid grid-cols-1 gap-x-6 gap-y-8 sm:grid-cols-6">
+            <div class="sm:col-span-4 space-y-4">
+            <label for="comments" class="block text-sm font-medium leading-6 text-gray-900">Comments</label>
```
holloway, line 111: Regarding `for="comments"` is this label for the following `<RpcTextarea>`? Will `<RpcTextarea>` have the correct DOM id "comments"

PR #479, opened 2025-09-12
feat: composable to load documentsComments

jennifer-richards:
Adds a composable to manage retrieving comments via the `documentsComments` API. The purpose is to ensure that pages making `useAsyncData()` calls with a shared key use the same handler and other parameters that must agree. See https://nuxt.com/docs/4.x/api/composables/use-async-data#shared-state-and-option-consistency for details.


rudimatz:
these changes are included in https://github.com/ietf-tools/purple/pull/478, I'll close this @jennifer-richards

PR #499, opened 2025-09-19
chore: refactor queue

holloway:
## chore

* Refactor queue with [tanstack table](https://tanstack.com/table/latest/docs/framework/vue/)
  * Style taken from existing table, refactored into granular components `<RpcTable>`, `<RpcThead>`, `<RpcTh>`, `<RpcTd>` etc so that Tanstack can use it.
  * Each tab has its own page. This makes the page logic easier to follow as a dev, with the downside of slightly less DRY (wet? whatever) in favour of clarity. The API calls are fewer and specific to each route, and the types of the api responses are typed in the table templates / filtering etc.
  * Sorting of dates works now
  * Tanstack is good and will allow us to do very complex tables

**screenshots**

Screenshot_2025-09-22_17-47-13

note the sorting arrows
Screenshot_2025-09-22_17-49-32


client/app/pages/queue/enqueuing.vue
```diff
@@ -0,0 +1,149 @@
...
+const currentTab: TabId = 'enqueuing'
+
+const {
+  data,
+  pending,
+  status,
+  refresh,
+  error,
+} = await useAsyncData(
+  `queue2-enqueuing`,
```
jennifer-richards, line 79: I don't know where we stand on this as lint, but this doesn't need to be a template literal since it's not interpolated.

client/app/pages/queue/queue.vue
```diff
@@ -0,0 +1,368 @@
...
+  state: {
+    get globalFilter() {
+      return JSON.stringify([needsAssignmentTristate.value, hasExceptionTristate.value, selectedLabelFilters.value])
+    },
+    get sorting() {
+      return sorting.value
+    },
+  },
+  globalFilterFn: (row) => {
+    console.log("'FILTER'")
```
jennifer-richards, line 308: debug statement

jennifer-richards (approved):
A couple minor comments, but overall looks good to me.

rudimatz (approved):

jennifer-richards (approved):

PR #504, opened 2025-09-24
feat: show consensus on draft info pages

rudimatz:
receive consensus value from DT and display it for import detail and queue detail page

use caching in memcache


rudimatz:
depends on https://github.com/ietf-tools/datatracker/pull/9605

rudimatz:
changes applied @jennifer-richards

PR #506, opened 2025-09-25
fix: restore previously reverted change of icon

rudimatz:
image


rudimatz:
@holloway you reverted this change, not use if it was on purpose

holloway:
@rudimatz Sorry, that wasn't on purpose. Probably was missed during a review, but also branching from an outdated version on xfer

holloway (approved):

Issue #507, opened 2025-09-25
queue UI improvements


rudimatz:
~~1. add RFC Number column DONE~~
~~2. remove status column DONE~~
~~3. make cluster sortable DONE~~
~~4. indicate columns that are sortable~~
~~5. disable sorting where it doesn't make sense (e.g. assignee, label)~~
~~6. add "weeks in queue" column~~
~~7. enable search (by draft name, rfc number ....) DONE~~
8. enable filters (label, current activity)

rudimatz:
8) addressed in https://github.com/ietf-tools/purple/pull/527

PR #508, opened 2025-09-25
feat: batch #1 UI improvements queue

rudimatz:
addresses some improvements from https://github.com/ietf-tools/purple/issues/507

- show RFC number and pages on queue page (`Pages` was previously mis-titled as `Status`)
- enable search for draft title and rfc number

### DIFF
--- a/client/app/pages/queue/queue.vue
+++ b/client/app/pages/queue/queue.vue
@@ -284,10 +284,7 @@ const columns = [
       sortingFn: (rowA, rowB, columnId) => {
         const a = rowA.getValue(columnId)?.number
         const b = rowB.getValue(columnId)?.number
-        if (a == null && b == null) return 0
-        if (a == null) return 1
-        if (b == null) return -1
-        return Number(a) - Number(b)
+        return (a > b) ? 1 : (a < b) ? -1 : 0
       },
       sortUndefined: 'last',
     }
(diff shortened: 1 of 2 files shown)


client/app/pages/queue/queue.vue
```diff
@@ -276,10 +277,17 @@ const columns = [
...
         ])
       },
-      sortingFn: 'alphanumeric',
+      sortingFn: (rowA, rowB, columnId) => {
+        const a = rowA.getValue(columnId)?.number
+        const b = rowB.getValue(columnId)?.number
+        if (a == null && b == null) return 0
+        if (a == null) return 1
+        if (b == null) return -1
+        return Number(a) - Number(b)
```
jennifer-richards, line 290: The [tanstack table docs for custom sorting functions](https://tanstack.com/table/latest/docs/guide/sorting#custom-sorting-functions) want the sorting fn to return `-1`, `0`, or `1`. I imagine what you have works, but for compliance with the spec maybe use:
```javascript
        return (a > b) ? 1 : (a < b) ? -1 : 0
```
(possibly with the `Number()` casts if they're needed)

jennifer-richards (changes requested):
Minor change to comply with tanstack docs requested

client/app/pages/queue/queue.vue
rudimatz: @jennifer-richards works fine with your solution, its certainly cleaner

jennifer-richards (approved):

PR #514, opened 2025-10-02
feat: show submitted in queue

rudimatz:


rpc/serializers.py
```diff
@@ -324,8 +325,21 @@ class Meta:
             "pending_activities",
             "rfc_number",
             "pages",
+            "submitted_at",
         ]
 
+    @extend_schema_field(serializers.DateField())
+    def get_submitted_at(self, obj):
```
jennifer-richards, line 332: This introduces an N+1 operation to the queue retrieval. It may be tolerable because N isn't enormous, but if it slows things down we should look at refactoring the `QueueViewSet` to use annotations / prefetch instead.

jennifer-richards (changes requested):
Requested change is the Math.floor comment, but I could be talked down from it. The other comment is just an observation to keep in mind if this slows things down perceptably.

rpc/serializers.py
rudimatz: I agree its a concern, but I think we should probably address it in separate, global refactoring

client/app/pages/queue/queue.vue
```diff
@@ -160,20 +160,45 @@ const columns = [
...
+    'submittedAt',
+    {
+      header: 'Submitted (Weeks in queue)',
+      cell: data => {
+        const value = data.getValue()
+
+        const submittedDate = DateTime.fromJSDate(value)
+        const now = DateTime.now()
+        const diffInDays = now.diff(submittedDate, 'days').days
+        const weeksInQueue = Math.round(diffInDays / 7 * 2) / 2 // Round to nearest 0.5
```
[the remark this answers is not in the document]
jennifer-richards: > @rjsparks can you confirm whether to show weeks as floor or with 0.5 precision

fwiw, I had in mind "floor AND with 0.5 precision". The math still works, it just doesn't bump to the next 0.5 until it's actually reached it.

jennifer-richards (approved):

PR #519, opened 2025-10-06
feat: enable debug toolbar

rudimatz:
enable BrowsableAPIRenderer like

http://localhost:8088/api/rpc/queue/
image

allow profiling in Debug toolbar (right side)

image

This toolbar will only appear on localhost (see `development.py`) and Django generated views, not the nuxt pages (so it will not affect purple UI)


rudimatz:
JSON output can be triggered in browser using http://localhost:8088/api/rpc/queue/?format=json

purple/settings/base.py
```diff
@@ -96,6 +96,7 @@
     ],
     "DEFAULT_RENDERER_CLASSES": [
         "rest_framework.renderers.JSONRenderer",
+        "rest_framework.renderers.BrowsableAPIRenderer",
```
jennifer-richards, line 99: Can we move this renderer config into `settings/development.py` also?

rudimatz: yes I think so

jennifer-richards (approved):

PR #541, opened 2025-10-13
feat: assignments finished

holloway:
## feat

* Allow assignments to be set to 'done' in a modal on the `/docs/[id]` route
* The time spent can be updated regardless of assignment state. The time spent is saved when the user browser focus leaves the text field.

**screenshot**
Screenshot_2025-10-21_16-03-10

### DIFF
--- a/client/app/components/AssignmentFinishedModalPerson.vue
+++ b/client/app/components/AssignmentFinishedModalPerson.vue
@@ -7,15 +7,14 @@
       <label class="text-xs">time spent:
         <input type="text" size="4" :id="props.assignment.id?.toString() ?? 'assignment'"
           v-model="props.assignment.timeSpent" class="text-xs p-1 bg-white text-black dark:bg-black dark:text-white"
-          :readonly="props.assignment.state === 'done'">
+          @blur="patchTimeSpent"
+        >
       </label>
       <BaseButton v-if="props.assignment.state !== 'done'" btnType="default" @click="finishAssignment" size="xs"
         :disabled="isSaving">
         Finish
       </BaseButton>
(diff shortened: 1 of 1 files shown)


rudimatz:
from BE perspective, you should be able to store the `time_spent` by sending a PUT or PATCH, even if it's done. However there is a bug on the PATCH, looking into it.
It might be useful to be able to re-open an assignment, we can discuss this later with RPC regarding their needs

holloway:
> Just for understanding, is there a backend-driven reason to re-open the assignment in order to adjust the time spent or is it to keep the UI logic simpler?

For me, I found it useful as UI feedback to see how the timeSpent was parsed by the server; so that I could potentially correct it because the format is `D HH:MM:SS` (I think) which isn't intuitive to me. But I'm not the end user so my intuition might be wrong. Potentially we want a different widget rather than a textbox but I thought this was a good first attempt and we could wait for user feedback before doing that.

holloway:
@rudimatz  If we could patch the time when the user leaves the time input field to show the parsing. That would be even better than the current UX.

Then we might not need to be able to reopen assignments :shrug:

rudimatz (approved):

rudimatz:
@holloway with https://github.com/ietf-tools/purple/pull/554 you can simply patch the time and leave it `DONE`

rjsparks:
+1 on not reopening assignments (later) - people should be able to edit metadata of closed assignments, and if someone is coming back to a doc to do more work, it should be a new assignment, not a reopen.

holloway:
_Done_

rudimatz (approved):

PR #549, opened 2025-10-16
feat: subseries

rudimatz:
- add model and migration for subseries `SubseriesMember`
- add subseries property to RfcToBe serializer like 
```
            "consensus": null,
            "subseries": "BCP 40"
        },
```       
- add endpoint that lists all subseries relations and allows to add, delete and patch

### DIFF
--- a/rpc/serializers.py
+++ b/rpc/serializers.py
@@ -6,6 +6,7 @@ from dataclasses import dataclass
 from itertools import pairwise
 
 from django.db import IntegrityError
+from django.db.models import QuerySet
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 from rest_framework.fields import empty
@@ -391,7 +392,7 @@ class SubseriesDoc:
     number: int
 
     @property
-    def documents(self) -> list[RfcToBe]:
+    def documents(self) -> QuerySet[RfcToBe]:
         return RfcToBe.objects.filter(
             subseriesmember__type__slug=self.type,
             subseriesmember__number=self.number,


rudimatz:
changes after recent discussion:
- added model `SubseriesType`
- added migration for bcp, std, fyi
- change constraint on `SubseriesMember`
- change RfcToBeSerializer to contain subseries obj

rpc/models.py
```diff
@@ -826,3 +826,34 @@ def __str__(self):
...
+        ]
+
+    def __str__(self):
+        return (
+            f"RfcToBe {self.rfc_to_be.id} is part of subseries {self.type.slug} "
+            f" {self.number}"
+        )
+
+
+class SubseriesType(Name):
```
jennifer-richards: For consistency with other `Name` subclasses this would be `SubseriesTypeName`.

rpc/models.py
```diff
@@ -826,3 +826,34 @@ def __str__(self):
...
         )
+
+
+class SubseriesMember(models.Model):
+    """Tracks which RFC belongs to which subseries and its number"""
+
+    rfc_to_be = models.ForeignKey(RfcToBe, on_delete=models.PROTECT)
+    type = models.ForeignKey("SubseriesType", on_delete=models.PROTECT)
+    number = models.PositiveIntegerField()
+    history = HistoricalRecords()
```
jennifer-richards: Will this history be useful? It will be somewhat rare for a `SubseriesMember` to be created when an RFC is put in a subseries, then deleted if it is taken out. What I think we'd want from history is from the other side of the relationship: "over time, what RFCs has this subseries document comprised?"

Or can we do this efficiently by searching the historical model and I've forgotten?

rpc/models.py
rudimatz: @jennifer-richards added `api/rpc/subseries/bcp9/` in latest commit

rpc/api.py
```diff
@@ -972,3 +977,94 @@ def upstream_search(
...
+            "type", "rfc_to_be", "rfc_to_be__draft"
+        ).all()
+
+    def retrieve(self, request, subseries_slug=None):
+        """Get all RfcToBe items in a specific subseries"""
+
+        # Parse subseries slug (e.g., "bcp123" -> type="bcp", number=123)
+        match = re.match(r"^([a-z]+)(\d+)$", subseries_slug.lower())
+
+        if not match:
```
jennifer-richards: Probably best to test `if match is None`

jennifer-richards (changes requested):
Looking good but needs some additional adjustments. See inline.

rpc/serializers.py
```diff
@@ -347,37 +347,49 @@ class SubseriesMemberSerializer(serializers.ModelSerializer):
...
-            "display_name": f"{type_slug.upper()} {number}",
-            "type": type_slug,
-            "number": number,
-            "rfc_to_be_ids": sorted(rfc_to_be_ids),
-            "rfc_count": len(rfc_to_be_ids),
-        }
+    members = serializers.ListField(child=serializers.IntegerField())
+    rfc_count = serializers.ReadOnlyField()
+    slug = serializers.ReadOnlyField()
+    display_name = serializers.ReadOnlyField()
```
jennifer-richards: Does the `ReadOnlyField` introspect the field types correctly when not used with a `ModelSerializer`? I suspect it needs to be `serializers.IntegerField(read_only=True)` (and similar for the other fields)

rpc/serializers.py
```diff
@@ -378,30 +378,61 @@ def get_slug(self, obj) -> str:
...
     type = serializers.CharField()
     number = serializers.IntegerField()
-    members = serializers.ListField(child=serializers.IntegerField())
-    rfc_count = serializers.IntegerField(read_only=True)
-    slug = serializers.CharField(read_only=True)
-    display_name = serializers.CharField(read_only=True)
+    documents = serializers.SerializerMethodField()
+    rfc_count = serializers.SerializerMethodField()
+    slug = serializers.SerializerMethodField()
+    display_name = serializers.SerializerMethodField()
```
jennifer-richards: The `SerializerMethodField` methods really ought to be properties on the `Subseries` dataclass (whatever it is called) rather than in the serializer. This will make it easier to use the nested serializers.

rpc/serializers.py
```diff
@@ -378,30 +378,61 @@ def get_slug(self, obj) -> str:
...
+        for doc in documents:
+            rfc_to_be = doc.rfc_to_be
+            draft = getattr(rfc_to_be, "draft", None) if rfc_to_be else None
+            if rfc_to_be and draft:
+                docs.append(
+                    {
+                        "name": draft.name,
+                        "rfc_number": rfc_to_be.rfc_number,
+                    }
+                )
```
jennifer-richards: After refactoring this method into a `@property` on the dataclass, it should just return the document queryset (or list, or whatever iterable is convenient) here.

jennifer-richards (changes requested):
I think this needs some further refactoring.

jennifer-richards (approved):

PR #559, opened 2025-10-28
feat: handle obsoletes and updates as RelatedDocs

rudimatz:
- add relation types in populate migration file
- update unique constraint on **RpcRelatedDocument** that was to strict. Add `relationship` to constraint to allow for different types of relations for same source/target


jennifer-richards (approved):
Looks good to me. I believe it will require tearing down and rebuilding the database to get the constraints right (or cleverly doing a manual reverse migration / forward migration coordinated with pulling in the changes).

PR #582, opened 2025-11-07
feat: final approval

rudimatz:
add final approval endpoint to retrieve, list, patch (set as approved), delete

add history for it


rudimatz:
expose endpoint like

- /api/rpc/documents/{draft_name}/final_approvals/
- /api/rpc/documents/{draft_name}/final_approvals/{final_approval_id}/

PR #605, opened 2025-11-12
feat: add filter on queue-list for pending_final_approval

rudimatz:
this PR adds logic to create and update `FinalApprovals` and filter docs with pending `FinalApprovals`

`/api/rpc/queue/?pending_final_approval=true`

to be used for view
image

### DIFF
--- a/rpc/serializers.py
+++ b/rpc/serializers.py
@@ -315,34 +315,15 @@ class MinimalRfcToBeSerializer(serializers.ModelSerializer):
 class FinalApprovalSerializer(serializers.Serializer):
     """Serialize final approval information for an RfcToBe"""
 
-    id = serializers.IntegerField()
-    rfc_to_be = MinimalRfcToBeSerializer()
+    id = serializers.IntegerField(read_only=True)
+    rfc_to_be = MinimalRfcToBeSerializer(read_only=True)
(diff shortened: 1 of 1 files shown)


rpc/serializers.py
```diff
@@ -306,6 +306,51 @@ class Meta:
...
+    body = serializers.CharField(required=False, allow_blank=True)
+    requested = serializers.DateTimeField(required=False)
+    approver = BaseDatatrackerPersonSerializer()
+    approved = serializers.DateTimeField(required=False, allow_null=True)
+    overriding_approver = BaseDatatrackerPersonSerializer(
+        required=False, allow_null=True
+    )
+
+    class Meta:
+        model = FinalApproval
```
jennifer-richards: Since this is `Serializer` rather than `ModelSerializer`, I don't think the `Meta` class does anything. I think you can just remove it, but it'd be worth thinking about whether there was some behavior of the `ModelSerializer` that you were counting on. (You may need to manually set the `read_only` flags on the fields above)

jennifer-richards (changes requested):
One change that I think is just minor cleanup unless there's something lurking.

rpc/serializers.py
rudimatz: ok, Meta removed

PR #618, opened 2025-11-18
feat: add table for unusable RFC numbers

rudimatz:
populate and also add history for unusable RFC numbers


rpc/models.py
```diff
@@ -326,6 +326,8 @@ class Meta:
     def __str__(self):
         return str(self.number)
 
+    history = HistoricalRecords()
```
jennifer-richards: style nit: should move up to be with the other fields ahead of the `__str__` method

PR #620, opened 2025-11-19
feat: Implement crossref DOI submission

### LINKED ISSUE #585: Register new RFC with crossref
Upload DOI data to crossref.org

API: https://www.crossref.org/documentation/register-maintain-records/direct-deposit-xml/https-post/
Current script: https://github.com/rfc-editor/rpcwebsite/blob/master/rfc-ed/bin/rfc2doi.py

kesara:
Fixes #585

### DIFF
--- a/purple/settings/base.py
+++ b/purple/settings/base.py
@@ -164,7 +164,7 @@ CROSSREF_API = os.environ.get(
 )
 CROSSREF_USER = os.environ.get("PURPLE_CROSSREF_USER", "user")
 CROSSREF_PASSWORD = os.environ.get("PURPLE_CROSSREF_PASSWORD", "friend")
-CROSSREF_TIMEOUT = 1  # in seconds
+CROSSREF_TIMEOUT = 30  # in seconds
 DOI_REGISTRANT = "RFC Editor"
 DOI_DEPOSITOR = "RFC Production Center for the RFC Editor"
 DOI_PREFIX = "10.17487"


purple/settings/base.py
```diff
@@ -157,3 +157,17 @@
 CELERY_BROKER_URL = os.environ.get("PURPLE_BROKER_URL", "amqp://mq/")
 CELERY_BROKER_CONNECTION_RETRY_ON_STARTUP = True
 CELERY_TASK_IGNORE_RESULT = True  # ignore results unless specifically enabled
+
+# Crossref / DOI
+CROSSREF_API = "https://test.crossref.org/servlet/deposit"
+CROSSREF_USER = "user"
+CROSSREF_PASSWORD = "friend"
```
jennifer-richards: I assume the api/user/password need to become a secret for production? If so, these should be pulled from the environment as `PURPLE_CROSSREF_API_URL`, `PURPLE_CROSSREF_USER` and `PURPLE_CROSSREF_PASSWORD`. We've been trying to do most of this config in `base.py` by using the environment if available and falling back to dev defaults, but it could go in `production.py` if the logic is different.

If `PURPLE_CROSSREF_API_URL` is set in the environment, then the absence of either of the other two settings should be an error. The production config should probably check that the API_URL is set. We'll need to be careful not to give staging the real secrets.

purple/settings/base.py
```diff
@@ -157,3 +157,17 @@
...
+CROSSREF_VERSION = "4.4.2"
```
jennifer-richards: Will changes to `CROSSREF_VERSION` require changing the xml generation code to match the change in schema? If so, I don't think this is a setting, it's a constant in the generation code.

purple/settings/base.py
```diff
@@ -157,3 +157,17 @@
...
+DOI_REGISTRANT = "RFC Editor"
+DOI_DEPOSITOR = "RFC Production Center for the RFC Editor"
+DOI_PREFIX = "10.17487"
+DOI_EMAIL = "webmaster@rfc-editor.org"
+DOI_URL = "https://www.rfc-editor.org/info/"
+AUTHOR_ORGS = [
```
jennifer-richards: `DOI_AUTHOR_ORGS`?

purple/settings/base.py
kesara: Fixed with b422deac5bdc21227d7654f2d7c1465f6e479fb8

requirements.txt
```diff
@@ -13,4 +13,5 @@ mozilla-django-oidc>=4.0.0
 psycopg2>=2.9.7
 pymemcache>=4.0.0
 python-json-logger>=2.0.7
+responses>=0.25.8
```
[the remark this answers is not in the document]
kesara: responses has an industry backing and community following[^1] and it's been used in author-tools for ages. I haven't experienced any issues. IMO DT should move to responses too.

[^1]: https://github.com/getsentry/responses

purple/settings/base.py
kesara: I didn't named it with `DOI` prefix because it's not a DOI specific setting, but it makes sense to have that because this organization list will be used only for that purpose.

purple/settings/base.py
kesara: f3d11b34e2280ec73264594f0982c36ce49c507d introduces ENV lookup.

PR #628, opened 2025-11-20
refactor: remove body from finalapproval

rudimatz:
and change constraint to make approver a required field

### DIFF
--- a/rpc/models.py
+++ b/rpc/models.py
@@ -530,15 +530,6 @@ class FinalApproval(models.Model):
         else:
             return f"request for final approval from {self.approver}"
 
-    class Meta:
-        constraints = [
-            models.CheckConstraint(
-                check=models.Q(approver__isnull=False),
-                name="finalapproval_approver_required",
-                violation_error_message="approver is required",
-            ),
-        ]
-
 
 class ActionHolderQuerySet(models.QuerySet):
     def active(self):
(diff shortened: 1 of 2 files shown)


rpc/models.py
```diff
@@ -531,32 +528,14 @@ def __str__(self):
...
-                name="finalapproval_approval_override_requires_approver",
-                violation_error_message="approval override requires an approver be set",
-            ),
-            models.CheckConstraint(
-                check=(models.Q(body="") | models.Q(overriding_approver__isnull=True)),
-                name="finalapproval_body_approval_no_override",
-                violation_error_message="body approval cant be overridden",
+                check=models.Q(approver__isnull=False),
+                name="finalapproval_approver_required",
+                violation_error_message="approver is required",
```
jennifer-richards: This is redundant - removing the `null=True` from the `approver` field guarantees that this constraint is satisfied.

jennifer-richards (changes requested):
A bit of redundant code I think.

Rewriting old migrations rather than adding new ones will cause trouble for existing dev environments. If we're going to require tearing them down and starting over, we might as well do a completely clean migration regeneration. (If you did that and we're still at 21 migrations then we should look into why...)

rpc/models.py
rudimatz: removed it

PR #633, opened 2025-11-21
fix: Make existing tests error free

kesara:
This is a crude attempt at making existing tests work.

```
dev ➜ /workspace$ ./manage.py test
Found 12 test(s).
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
............
----------------------------------------------------------------------
Ran 12 tests in 0.083s

OK
Destroying test database for alias 'default'...
```

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -43,7 +43,6 @@ from rules.contrib.rest_framework import AutoPermissionViewSetMixin
 
 from datatracker.models import DatatrackerPerson, Document
 from datatracker.rpcapi import with_rpcapi
-
 from rpc.models import (
     ASSIGNMENT_INACTIVE_STATES,
     ActionHolder,
--- a/rpc/serializers.py
+++ b/rpc/serializers.py
@@ -15,7 +15,6 @@ from simple_history.utils import update_change_reason
 
 from datatracker.models import DatatrackerPerson, Document
 from datatracker.utils import build_datatracker_url
-
 from rpc.models import (
     ActionHolder,
     Assignment,
--- a/rpc/signals.py
+++ b/rpc/signals.py
@@ -9,7 +9,6 @@ from rpc.models import (
     ActionHolder,
     Assignment,
     ClusterMember,
-    DumpInfo,
     RfcToBe,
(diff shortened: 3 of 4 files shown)


jennifer-richards (commented):
I don't think this PR is on the right track, but I might not be understanding something about your test arrangement. When I run the tests on main right now 11 of 12 pass, and I think #646 cleans up the one that fails for me.

The relative import change should not be required. If the relative imports aren't working, I'm pretty sure that means there's something wrong with the python environment not with the code.

Django's test framework should automatically handle switching to a separate database. Barring some specific need, we want to let it do it for us - it creates it and tears it down as needed, guaranteeing we don't get cross-talk between separate test runs. [See the docs on this topic.](https://docs.djangoproject.com/en/5.2/topics/testing/overview/#the-test-database)

Confirming locally by putting a breakpoint in a test:
```
dev ➜ /workspace $ ./manage.py test
Found 12 test(s).
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
....> /workspace/rpc/tests.py(12)test_next_rfc_number()
-> self.assertEqual(next_rfc_number(), [1])
(Pdb) interact
*interactive*
>>> from django.conf import settings
>>> settings.DATABASES
{'default': {'ENGINE': 'django.db.backends.postgresql', 'NAME': 'test_rpctools', 'USER': 'rpc', 'PASSWORD': 'abcd1234', 'HOST': 'db', 'PORT': 5432, 'ATOMIC_REQUESTS': False, 'AUTOCOMMIT': True, 'CONN_MAX_AGE': 0, 'CONN_HEALTH_CHECKS': False, 'OPTIONS': {}, 'TIME_ZONE': None, 'TEST': {'CHARSET': None, 'COLLATION': None, 'MIGRATE': True, 'MIRROR': None, 'NAME': None}}}
>>>
now exiting InteractiveConsole...
(Pdb) c
.E......
======================================================================
ERROR: test_default_factories_dont_crash (rpc.tests_factories.BasicRpcFactoryTests.test_default_factories_dont_crash)
```
(note in particular `'NAME': 'test_rpctools'`)

PR #634, opened 2025-11-21
feat: remove logic to delete

rudimatz:
not issued RFC numbers are blocked forever, no need to be able to delete them


jennifer-richards (commented):
I'm puzzled about the motivation to remove this capability. Even if "not issued" is forever, users make mistakes and this is taking away the ability to fix that without admin intervention. Seems like it's asking for extra work later.

jennifer-richards (approved):

PR #641, opened 2025-11-25
ci: Run django tests

### LINKED ISSUE #640: Run django tests as a PR check
Run django tests as part of PR checks/test GHA.[^1]

[^1]: https://github.com/ietf-tools/purple/blob/main/.github/workflows/test.yml

kesara:
Fixes #640 

This PR is depended on #646 to pass the tests.


jennifer-richards:
Maybe "lint-checks" or "python-lint" instead of "ruff-tests"? Not sure, ruff is accurate I guess.

For the other, I think `django-tests` is fairly logical - it's the tests that Django's `manage.py test` runs.

If we've already got names for these at dt, we should match them, tho.

rudimatz:
relates to https://github.com/ietf-tools/purple/pull/646

PR #650, opened 2025-11-26
feat: approval logs (public comments)

holloway:
## feat

* approval logs, based off the components for comments. I've forked the comment components rather than abstracting/generalising them to keep it simpler, allow them to deviate in the future etc. If in the future we want to  consolidate the components we can.
* for comments (private) and approval logs (public) I've added 'private' and 'public' suffixes to headings and buttons to make it very clear to users. See screenshot

*screenshot*
Screenshot_2025-11-27_11-03-55


holloway:
@rudimatz the screenshot of 'comments' is just to show a UI text change of a 'private' suffix in their heading/button, and that approval logs have a 'public' suffix in their heading/button. So the screenshot of 'comments' showing 'system' isn't a change in this PR

PR #684, opened 2025-12-03
feat: create and update cluster and order

rudimatz:
endpoints to create and update cluster; functions to add and remove single drafts from cluster and reorder

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -590,7 +590,6 @@ class ClusterViewSet(
 
     @extend_schema(
         operation_id="clusters_remove_document",
-        request=ClusterAddRemoveDocumentSerializer,
         responses=ClusterSerializer,
         examples=[
             OpenApiExample(
@@ -637,7 +636,6 @@ class ClusterViewSet(
 
     @extend_schema(
         operation_id="clusters_reorder_documents",
-        request=ClusterReorderDocumentsSerializer,
         responses=ClusterSerializer,
         examples=[
             OpenApiExample(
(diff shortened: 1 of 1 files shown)


rpc/api.py
```diff
@@ -509,16 +510,184 @@ class CapabilityViewSet(viewsets.ReadOnlyModelViewSet):
...
+        except Document.DoesNotExist:
+            return Response(
+                {"error": f"Document with name '{draft_name}' not found"},
+                status=status.HTTP_404_NOT_FOUND,
+            )
+
+        # Check if document is already in any cluster
+        existing_member = ClusterMember.objects.filter(doc=doc).first()
+        if existing_member:
+            return Response(
```
jennifer-richards: See above re: raising exceptions. If you squint this is still a validation error

rpc/api.py
```diff
@@ -519,6 +527,165 @@ def get_queryset(self):
...
+        ClusterMember.objects.create(cluster=cluster, doc=doc, order=max_order + 1)
+
+        cluster.refresh_from_db()
+
+        serializer = self.get_serializer(cluster)
+        return Response(serializer.data)
+
+    @extend_schema(
+        operation_id="clusters_remove_document",
+        request=ClusterAddRemoveDocumentSerializer,
```
jennifer-richards: suggestion above about moving the request serializer into the `@action()` applies here as well

jennifer-richards (approved):

PR #708, opened 2025-12-09
feat: add fields to cluster API payload

rudimatz:
add field `is_active` on Cluster
add fields `is_received`, `references` and `disposition` in ClusterMembers
use annotation to avoid n+1

add filtering (active/inactive) and ordering for clusters

### DIFF
--- a/rpc/serializers.py
+++ b/rpc/serializers.py
@@ -5,6 +5,7 @@ import warnings
 from dataclasses import dataclass
 from itertools import pairwise
 
+import rpcapi_client
 from django.db import IntegrityError, transaction
 from django.db.models import QuerySet
 from drf_spectacular.utils import extend_schema_field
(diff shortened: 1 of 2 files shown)


rpc/serializers.py
```diff
@@ -847,6 +854,63 @@ def get_rfc_number(self, clustermember: ClusterMember) -> int | None:
...
+        rfctobe = getattr(doc, "rfctobe_set", None)
+        if rfctobe is not None:
+            rfctobe = rfctobe.exclude(disposition__slug="withdrawn").first()
+
+        if not rfctobe:
+            return None
+
+        return rfctobe.disposition.slug
+
+    def get_references(self, clustermember: ClusterMember) -> list[dict] | None:
```
jennifer-richards: Does this annotation result in a valid schema / API client types? I'm not sure what `list[dict]` will turn into when spectacular looks at it.

PR #724, opened 2025-12-14
fix: styling; feat: prevent rfc # edit of published rfcs

holloway:
## feat

* if RFC is published (if it has a `.publishedAt` date) then don't show RFC number editing UI

## fix

* table sort buttons (previously ↕ was shown regardless)
* dark mode across queue/* and doc info routes (badges, basecard)

**screenshots**

*before*
Screenshot_2025-12-15_13-21-07
*after*
Screenshot_2025-12-15_13-18-31

*before*
Screenshot_2025-12-15_13-21-13
*after*
Screenshot_2025-12-15_13-21-22

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,7 +98,7 @@ jobs:
           cp purple_api.yaml ..
 
       - name: Upload API
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: api-clients
           path: |
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -364,6 +364,27 @@ def import_submission(request, document_id, rpcapi: rpcapi_client.PurpleApi):
         with transaction.atomic():
             rfctobe = serializer.save()
 
+            # check for existing references where the new draft is the target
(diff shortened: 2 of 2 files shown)


client/app/components/DocInfoCard.vue
```diff
@@ -52,8 +52,11 @@
         <DescriptionListItem term="Disposition" :details="rfcToBe.disposition" :spacing="spacing" />
         <DescriptionListItem term="RFC Number" :spacing="spacing">
           <DescriptionListDetails>
-            <div v-if="!props.isReadOnly" class="flex items-center gap-2">
-              <EditRfcNumber :name="rfcToBe.name" :initial-rfc-number="rfcToBe.rfcNumber" :on-success="() => props.refresh()" />
+            <div v-if="!props.isReadOnly &&
+              // published RFCs can't be edited
+              !rfcToBe.publishedAt" class="flex items-center gap-2">
```
rudimatz: The cleaner way to check is via the `disposition`. Check if rfctobe.disposition = 'published'

rudimatz (changes requested):

Issue #738, opened 2025-12-17
Cluster graph shows queued I-Ds as not received


ajeanmahoney:
For example, the graph for Cluster 517 shows that draft-ietf-teas-ietf-network-slice-nbi-yang and draft-ietf-ccamp-rfc9093-bis are not received, but purple lists them in the queue:

<img width="1106" height="148" alt="Image" src="https://github.com/user-attachments/assets/7e69fde1-8d75-4bba-bc16-6bcb28b642d5" />

<img width="811" height="738" alt="Image" src="https://github.com/user-attachments/assets/67af6a17-a2cf-4230-9a13-243bd8cb477d" />

ajeanmahoney:
Debug info has the following:
```
{
  "number": 517,
  "documents": [
    {
      "name": "draft-ietf-ccamp-layer1-types",
      "disposition": "in_progress",
      "references": [
        {
          "id": 24,
          "relationship": "not-received",
          "draftName": "draft-ietf-ccamp-layer1-types",
          "targetDraftName": "draft-ietf-teas-rfc8776-update"
        }
      ],
      "isReceived": true,
      "order": 1
    },
    {
      "name": "draft-ietf-teas-rfc8776-update",
      "isReceived": false,
      "order": 2
    },
    {
      "name": "draft-ietf-ccamp-l1csm-yang",
      "disposition": "in_progress",
      "references": [
        {
          "id": 28,
          "relationship": "refqueue",
          "draftName": "draft-ietf-ccamp-l1csm-yang",
          "targetDraftName": "draft-ietf-ccamp-layer1-types"
        },
        {
          "id": 107,
          "relationship": "not-received-2g",
          "draftName": "draft-ietf-ccamp-l1csm-yang",
          "targetDraftName": "draft-ietf-teas-rfc8776-update"
        }
      ],
      "isReceived": true,
      "order": 3
    },
    {
      "name": "draft-ietf-teas-applicability-actn-slicing",
      "disposition": "in_progress",
      "references": [
        {
          "id": 31,
          "relationship": "not-received",
          "draftName": "draft-ietf-teas-applicability-actn-slicing",
          "targetDraftName": "draft-ietf-teas-actn-pm-telemetry-autonomics"
        },
        {
          "id": 32,
          "relationship": "not-received",
          "draftName": "draft-ietf-teas-applicability-actn-slicing",
          "targetDraftName": "draft-ietf-teas-te-service-mapping-yang"
        },
        {
          "id": 33,
[...]

PR #760, opened 2026-01-08
fix: drafts in graph not always properly tagged

rudimatz:
instead of using `uniq` and using the first of multiple elements with same ID, I rather want to merge the properties of all those elements

fixes issue of e.g. https://purple.staging.ietf.org/clusters/562/, where `draft-ietf-bfd-optimizing-authentication` is received, but displayed as not received

image

fixes https://github.com/ietf-tools/purple/issues/738

### DIFF
--- a/client/app/components/DocumentDependenciesGraph.vue
+++ b/client/app/components/DocumentDependenciesGraph.vue
@@ -124,14 +124,16 @@ const setTooltip: SetTooltip = (props) => {
 const hasMounted = ref(false)
 
 const rfcsByDraftName = computed(() => {
-  const result: Record<string, RfcToBe> = {};
+  const result: Record<string, RfcToBe> = {}
   if (props.rfcsToBe) {
     for (const rfcToBe of props.rfcsToBe) {
-      if (rfcToBe.name) result[rfcToBe.name] = rfcToBe;
+      if (rfcToBe.name) {
+        result[rfcToBe.name] = rfcToBe
+      }
     }
   }
-  return result;
-});
+  return result
+})
 
 const clusterGraphData = computed(() => {
 


jennifer-richards:
Just to be clear, what I'm wondering is whether we're going to have cases like
```
nodes = [
  {"id": "1", "property_a": "correct value"},  # what was picked before
  {"id": "1", "property_a": "incorrect value", "property_b": "value"},  # picked now, overwriting property_a
]
```
cropping up, where we're just immediately introducing a new bug. If you've looked at a few cases to see that there aren't side effects like this and we need feedback from the RPC / wider testing to evaluate, then I don't object to the change here. Just want to be sure this patch isn't too narrowly focused on fixing a specific case.

client/app/components/DocumentDependenciesGraph.vue
```diff
@@ -123,7 +123,23 @@ const setTooltip: SetTooltip = (props) => {
...
 
+const rfcsByDraftName = computed(() => {
+  const result: Record<string, RfcToBe> = {};
+  if (props.rfcsToBe) {
+    for (const rfcToBe of props.rfcsToBe) {
+      if (rfcToBe.name) result[rfcToBe.name] = rfcToBe;
+    }
+  }
+  return result;
+});
```
jennifer-richards: Style
```suggestion
const rfcsByDraftName = computed(() => {
  const result: Record<string, RfcToBe> = {}
  if (props.rfcsToBe) {
    for (const rfcToBe of props.rfcsToBe) {
      if (rfcToBe.name) {
        result[rfcToBe.name] = rfcToBe
      }
    }
  }
  return result
})
```

jennifer-richards (approved):
lgtm

fwiw, the one-liner `if (rfcToBe.name)...` that I asked you to break up and add braces was ok as you had it, according to StandardJS style. My bad. If you want to change it back feel free.

PR #772, opened 2026-01-14
feat: use celery to retrieve metadata from XML in repo

rudimatz:
create endpoint to create a celery task to retrieve XML from given repo and extract metadata from it. Then it will store it in DB.

A separate call will check if the task is finished and if yes, will return the extracted metadata from the DB

Enable celery backend to be able to get status of individual tasks (and check if they are finished)

Prevent unpredictable behavior in race conditions by refusing a 2nd task for same RfcToBe if a previous task is still in-flight.


rpc/lifecycle/repo.py
```diff
@@ -133,10 +134,78 @@ def get_file(self, path: PurePath | str) -> GithubRepositoryFile:
...
 
 class RepositoryError(Exception):
     """Base class for repository exceptions"""
 
 
 class TemporaryRepositoryError(RepositoryError):
     """Repository exception that is likely temporary and worth retrying"""
+
+
+class Metadata:
```
jennifer-richards: Is this class going to contain more than a single static method soon? If not, I'd suggest dropping it and just having a plain parse function instead.

PR #789, opened 2026-01-21
fix: add config for devcontainers celery

rudimatz:
currently, celery container fails to start because of permission issue when using devcontainers.
This change will create a separate image for the celery container and set the proper file permissions (and allows for custom overwrite)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ __pycache__
 
 /rpcapi.yaml
 /openapi/rpcapi_client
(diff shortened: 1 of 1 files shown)


jennifer-richards (commented):
Seems ok to me if it works, but I still don't understand why this is needed for the celery container but not for the app container. I'll let @NGPixel (and maybe @rjsparks ?) weigh in as far as approving.

NGPixel (approved):

PR #791, opened 2026-01-21
feat: add sync logic

rudimatz:
this PR will add a `/sync` to the existing endpoint.

the underlying `update_metadata` function has been refactored to loop though the metadata fields and update **only** those fields that are actually different.

### DIFF
--- a/rpc/tests_serializers.py
+++ b/rpc/tests_serializers.py
@@ -77,6 +77,9 @@ class SerializerTests(TestCase):
                             "to Network-Based Security Services"
                         ),
                         "is_match": True,
+                        "can_auto_fix": False,
+                        "is_error": False,
+                        "detail": "",
                     },
                 },
                 {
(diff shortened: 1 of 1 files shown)


rpc/lifecycle/metadata.py
```diff
@@ -93,114 +93,193 @@ def parse_rfc_xml(xml_string):
...
+                if not is_match and can_fix:
+                    if field == "title":
+                        new_title = metadata.get("title")
+                        if new_title:
+                            rfctobe.draft.title = new_title
+                            rfctobe.draft.save(update_fields=["title"])
+                            updated_fields["title"] = new_title
+
+                    elif field == "publication_date":
+                        pub_date = metadata.get("publication_date")
```
jennifer-richards: Are the publication dates always full dates? Drafts often contain only the year or year + month.

If they're always full dates for modern RFCs, we should consider refactoring just to use an ISO date in the metadata field (i.e., instead of `{"month":...}` just store `"2026-01-01"` and simplify parsing / reconstructing.

If they're not always full dates, we should clarify what's valid and handle it appropriately.

Issue #808, opened 2026-01-26
Model the consensus boilerplate field


jennifer-richards:
The RPC tracks whether an RFC should use the "consensus boilerplate" using a field called `consensus`. This is currently modeled via a pass-through to`datatracker.Document.consensus`, which fetches the value from datatracker for the RFC's progenitor draft. The datatracker is not always correct about the status, and it occasionally changes during the editing process. Need to add a `consensus` boolean field to `RfcToBe`. This should be initialized to the value from datatracker during the `import_submission()` process. The value should replace the existing pass-through, and should be exposed for update via the `partialUpdate` API. 

This should be a pretty small change.

PR #819, opened 2026-01-27
feat: add consensus field on RfcTobe

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/808

makes field patchable

TODO: update the consensus field from DT in xfer


jennifer-richards (approved):
In addition to the xfer change, also need to import the draft value during document ingestion.

PR #823, opened 2026-01-28
feat: change block assignment logic

rudimatz:
when a blocking condition arises, close current active assignments, create a new one for each of those with state blocked.

If the blocking condition goes away, close the blocked assignments, and create a new one for each.

UI: add a ordering feature for assignments to make it more intuitive.

### DIFF
--- a/rpc/lifecycle/blocked_assignments.py
+++ b/rpc/lifecycle/blocked_assignments.py
@@ -185,12 +185,14 @@ def _create_blocked_assignments(rfc: RfcToBe, reasons: set[str]) -> bool:
                 comment = (
                     f"blocked because of blocking condition(s): {', '.join(reasons)}; "
                 )
-                Assignment.objects.get_or_create(
+                Assignment.objects.update_or_create(
                     rfc_to_be=rfc,
                     role=role,
-                    comment=comment,
                     person=assignment.person,
                     state=Assignment.State.IN_PROGRESS,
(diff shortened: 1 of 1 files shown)


client/app/pages/docs/[id]/assignments.vue
```diff
@@ -98,8 +98,28 @@ const { data: assignments, refresh: refreshAssignments } = await useAsyncData(
...
+const assignmentOrder = [
+  'enqueuer',
+  'formatting',
+  'ref_checker',
+  'first_editor',
+  'second_editor',
+  'final_review_editor',
+  'publisher',
+  'blocked'
+]
```
Comment (jennifer-richards) [inline client/app/pages/docs/[id]/assignments.vue]: Should perhaps become an order field on `RpcRole` eventually

rpc/lifecycle/blocked_assignments.py
```diff
@@ -230,9 +227,41 @@ def _close_latest_blocked_assignment(rfc: RfcToBe) -> bool:
...
+                ):
+                    latest_assignment = assignment
+                    latest_history_date = hist.history_date
+            if latest_assignment:
+                logger.info(
+                    "Creating new assignment for last closed_for_hold for rfc %s and person %s",
+                    rfc.pk,
+                    a.person,
+                )
+                Assignment.objects.create(
```
jennifer-richards: Again with the `get_or_create()` suggestion

PR #856, opened 2026-02-09
fix: changes to make the vitest succeed

rudimatz:


holloway (approved):
LGTM just delete that other line too

PR #865, opened 2026-02-11
feat: ignore whitespaces in abstract compare

rudimatz:
as well as tabs, newlines etc


ajeanmahoney:
Keeping paragraph breaks would be helpful

PR #871, opened 2026-02-11
fix: remove client from git, build client for tests in GH workflow

rudimatz:
as discussed, we remove the purple client from git, but add the yml file


holloway (approved):

.github/workflows/website.yml
```diff
@@ -28,10 +28,16 @@ jobs:
         run: npm ci
         working-directory: ./client
 
+      - name: Generate API client
+        run: |
+          npx --yes @openapitools/openapi-generator-cli generate --generator-key purple
+          sed -i '1i // @ts-nocheck' client/app/purple_client/runtime.ts
```
holloway: if we use a later version of the generator (as you say v7.19 from 7.1) doesn't it remove the need for this?

Issue #864, opened 2026-02-11
Set RfcToBe.repository via UI workflow


jennifer-richards:
Need UI / workflow steps to set/edit the `repository` field for an `RfcToBe`

jennifer-richards:
> and the default would be `https://github.com/rfc-editor-drafts/{draft_name}` ?

This should become a variable in settings, in case that wasn't obvious.

NGPixel:
That URL is correct.

jennifer-richards:
There's a UI for setting the repository name, but not for populating it on ingestion (either automatically or manually) that I can see.

PR #884, opened 2026-02-16
feat: add repo in DocInfo UI

rudimatz:
image


rudimatz:
addresses https://github.com/ietf-tools/purple/issues/864

PR #889, opened 2026-02-18
chore(dev): untangle api build / deps

jennifer-richards:


docker/scripts/app-init.sh
```diff
@@ -35,25 +42,20 @@ echo "Populating initial history..."
...
-
 # Django should be operational now. Build purple API client.
 ./manage.py spectacular --file purple_api.yaml && \
-    client/node_modules/.bin/openapi-generator-cli generate --generator-key purple  || true
-    echo "If not set, add @ts-nocheck in runtime.ts to avoid type errors from generated code"
-    if ! grep -q "// @ts-nocheck" "client/app/purple_client/runtime.ts"; then
-        sed -i '1i // @ts-nocheck' "client/app/purple_client/runtime.ts"
-    fi
-    /bin/cp purple_api.yaml client/app/purple_client/.purple_api.yaml
+    npx --yes @openapitools/openapi-generator-cli@2.28 generate --generator-key purple  || true
```
kesara: Line 15 also installs `@openapitools/openapi-generator-cli@2.28`

jennifer-richards: It's just specifying it everywhere that it uses it. We might be able to relax that, but until we straighten everything out I don't want there to be any ambiguity

PR #894, opened 2026-02-18
feat: Add published formats model

kesara:


rpc/models.py
```diff
@@ -272,6 +272,8 @@ class IanaStatus(models.TextChoices):
         help_text="Whether document has consensus (None=unknown)",
     )
 
+    formats = models.ManyToManyField("PublishedFormatName", blank=True)
```
rudimatz: maybe better to call in published_formats to be more explicit

jennifer-richards (approved):

PR #896, opened 2026-02-18
feat: notifications on change

rudimatz:
1. install celery beat and add history for all relevant models where missing
2. have a celery task run x minutes that checks if changes in history occurred in last 60 seconds (threshold)
3. compile list of RfcToBe IDs of changed records in queue (disposition active)
4. send the notification

- added a model `PeriodicTaskRun` to keep track of when task last ran and use that time as cutoff time for history changes at next valid task run

- default for task runner is 5 minutes, but can be changed in settings

### DIFF
--- a/rpc/tasks.py
+++ b/rpc/tasks.py
@@ -165,8 +165,8 @@ def process_rfctobe_changes_for_queue_task():
         process_rfctobe_changes_for_queue()
     except Exception as e:
         logger.error(f"Error in process_rfctobe_changes_for_queue_task: {e}")
-        
-        
+
+
 @shared_task(bind=True)
 def create_index(self):
     createRfcTxtIndex()


jennifer-richards (commented):
Just commenting for now to get the inline comments to post.

I woul like to see the celery tasks and "business logic" better separated but I think we should discuss that interactively rather than via PR comments.

rpc/tasks.py
```diff
@@ -152,3 +157,25 @@ class PublishRfcToBeTask(RetryTask):
 def publish_rfctobe_task(self, rfctobe_id, expected_head):
     rfctobe = RfcToBe.objects.get(pk=rfctobe_id)
     publish_rfctobe(rfctobe, expected_head=expected_head)
+
+
+class NotifyQueueTask(RetryTask):
+    max_retries = 10  # after that, manual intervention is likely needed
```
jennifer-richards: This will retry, but... we're planning to call this task periodically at a pace no more than every few minutes. It _might_ make sense to retry a couple times with a several second backoff just to avoid brief network failures or other interruptions, but it may cause problems if the retries last longer than that. If datatracker is down for an hour and we run the task, say, every minute, then at the end of the hour we'll have all the retries for 60 task runs queued up. That'll clog celery unnecessarily.

Also given the `RetryTask` implementation we have now, running out of retries will trigger an email to the admins. That is likely to result in a lot of emails in a short time. I think we should take out the retries for now and plan to handle failure reporting in the future.

rpc/lifecycle/notifications.py
```diff
@@ -0,0 +1,193 @@
...
+        else:
+            logger.info("No in-progress RFCs changed")
+
+        task_run.last_run_at = current_check_time
+        logger.info("Completed processing history changes")
+
+    except Exception as e:
+        logger.exception(f"Unexpected error in process_rfctobe_changes_for_queue: {e}")
+        task_run.is_running = False
+        task_run.save()
```
jennifer-richards: Suggest adding a `finally:` that does the reset of `is_running` rather than including it in both execution branches. That'll make it clearer that it _must_ happen.

PR #920, opened 2026-02-26
feat: celery in k8s should enable beat

rudimatz:
settings applied as in `errata`, https://github.com/search?q=repo%3Aietf-tools%2Ferrata%20beat&type=code


jennifer-richards (changes requested):
couple further adjustments needed

PR #921, opened 2026-02-26
feat: update published RFCs in DT

rudimatz:
for published RFCs, offer button to "sync" metadata changes through to DT

image

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -31,6 +31,7 @@ from rest_framework.decorators import (
     api_view,
 )
 from rest_framework.exceptions import (
+    APIException,
     NotAuthenticated,
     NotFound,
     PermissionDenied,
@@ -873,7 +874,7 @@ class RfcToBeViewSet(viewsets.ModelViewSet):
     @extend_schema(
         operation_id="documents_sync_metadata",
         request=None,
-        responses={200: None, 400: None},
+        responses={200: None},
     )
     @action(detail=True, methods=["post"], url_path="sync_metadata")
     @with_rpcapi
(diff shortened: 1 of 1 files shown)


rpc/api.py
```diff
@@ -870,6 +870,56 @@ def publish(self, request, draft__name=None):
...
+    @with_rpcapi
+    def sync_metadata(self, request, rpcapi: rpcapi_client.PurpleApi, draft__name=None):
+        """Push current RFC metadata to the datatracker via
+        rpcapi purple_rfc_partial_update."""
+        rfctobe = self.get_object()
+        if rfctobe.rfc_number is None:
+            return Response(
+                {"detail": "No RFC number assigned"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
jennifer-richards: I think raising a serializers.ValidationError here will use the standardized-errors tooling to return an error here in a way that matches other endpoints.

rpc/api.py
```diff
@@ -870,6 +870,56 @@ def publish(self, request, draft__name=None):
...
+            if rfctobe.publication_std_level
+            else None,
+            subseries=[
+                f"{m.type.slug}{m.number}" for m in rfctobe.subseriesmember_set.all()
+            ],
+        )
+        rpcapi.purple_rfc_partial_update(
+            rfc_number=str(rfctobe.rfc_number),
+            patched_editable_rfc_request=patched,
+        )
```
jennifer-richards: Probably should be looking at the return value or catching exceptions here

jennifer-richards (changes requested):
A couple comments. It's ok to put off checking the return values from the datatracker API call, but search for `serializers.ValidationError` in `api.py` to see what I mean about raising an error instead of manually constructing a 4xx response.

jennifer-richards (approved):
We'll want to do more work on the error handling - a 500 isn't really the right response to a foreseeable API error - but this is ok until we have a better idea what to do.

Issue #584, opened 2025-11-08
Ability to capture and update keywords for each document


ajeanmahoney:
The rfced database stores keywords that are used for search. Keywords will be used as tags on Red in the future. Some RFCs don't have keywords because the keywords have already been used in the title or abstract.

Editors should be able to add, remove, update keywords for any published RFCs or docs in the queue, but it is okay not to provide keywords for a document.

PR #925, opened 2026-02-27
feat: add keywords

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/584


rudimatz:
should be editable even AFTER publication

Issue #874, opened 2026-02-12
Editor assignment time not being stored


ajeanmahoney:
A user can add time to their assignment on the Finish Assignments page, and the state moves to Done, but the time is not recorded. That is, "time spent" is set back to 0 when the user tabs away and back. Nor is there a item added to the History tab.

PR #939, opened 2026-03-03
fix: saving time_spent for finishing assignments

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/874

currently the UI does 2 API calls, one overwriting the other


holloway (approved):
LGTM

Issue #981, opened 2026-03-12
Cluster 430: Queue page shows 6 docs, Cluster page shows 5 docs


ajeanmahoney:
In [production](https://www.rfc-editor.org/cluster_info.php?cid=C430) and on the purple Queue page, there are 6 documents in C430, which is the correct number. On the Cluster Management and [Cluster 430](https://purple.staging.ietf.org/clusters/430/) pages in purple, the diagram and list are missing [draft-ietf-tls-keylogfile](https://purple.staging.ietf.org/docs/draft-ietf-tls-keylogfile), which is in C430 because it has a normative ref to [draft-ietf-tls-rfc8446bis](https://purple.staging.ietf.org/docs/draft-ietf-tls-rfc8446bis). The info page for tls-keylogfile shows this correctly.

Issue #956, opened 2026-03-06
If a cluster contains only pub RFCs and I-Ds that have never entered the queue, it's inactive


ajeanmahoney:
For example, the Cluster Management page lists C238 as active but it only contains published RFCs and I-Ds that never entered the queue (TBF, we probably should have removed these I-Ds from the list of normative references of docs that were published)

PR #991, opened 2026-03-13
fix: handle multiple drafts per rfctobe for cluster

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/981, https://github.com/ietf-tools/purple/issues/956

needed better handling of possibility of multiple rfctobes per draft

### DIFF
--- a/rpc/serializers.py
+++ b/rpc/serializers.py
@@ -1162,7 +1162,8 @@ class ClusterSerializer(serializers.ModelSerializer):
         fields = ["number", "documents", "draft_names", "is_active"]
 
     def get_is_active(self, cluster) -> bool:
-        """A cluster is considered active if at least one of its documents is in_progress."""
+        """A cluster is considered active if at least one of its documents is
+        in_progress."""
 
         # Use annotated value if available
         if hasattr(cluster, "is_active_annotated"):


rpc/models.py
```diff
@@ -577,8 +577,12 @@ def with_data_annotated(self):
         return self.prefetch_related(
             Prefetch(
                 "clustermember_set",
-                queryset=ClusterMember.objects.exclude(
-                    doc__rfctobe__disposition__slug="withdrawn"
+                queryset=ClusterMember.objects.filter(
+                    Exists(
+                        RfcToBe.objects.filter(draft=OuterRef("doc")).exclude(
+                            disposition__slug="withdrawn"
+                        )
```
jennifer-richards: Curious - would 
```python
ClusterMember.objects.filter(doc__rfctobe__disposition__slug="in_progress")
```
do here? Or possibly with `..._slug_in=["in_progress", "created"]` or something?

rpc/models.py
```diff
@@ -614,9 +618,9 @@ def with_is_active_annotated(self):
         """
         return self.annotate(
             is_active_annotated=Exists(
-                ClusterMember.objects.filter(cluster=OuterRef("pk"))
-                .exclude(doc__rfctobe__disposition__slug="published")
-                .exclude(doc__rfctobe__disposition__slug="withdrawn")
+                RfcToBe.objects.filter(draft__clustermember__cluster=OuterRef("pk"))
+                .exclude(disposition__slug="published")
+                .exclude(disposition__slug="withdrawn")
```
jennifer-richards: Similarly, maybe filter over `slug__in=["in_progress", "created"]`? (or just `in_progress`, I'm not sure if `created` should count.

PR #998, opened 2026-03-14
chore: removing assumption about user id

rudimatz:
cleanup: remove assumption in config about local user's UID/GID. Instead provide example how to set it up

The file permission issue prevents celery container to properly start when using VS Code's DevContainers


holloway (approved):
WFM

PR #1010, opened 2026-03-18
feat: improve dialog for adding dependencies

rudimatz:
improve the dialog to add Publishing dependencies on a draft's individual info page.
Include a search functionality (with filter to only return `in_progress` docs).
Filter added on backend side. If result found and selected, enable button to add as "refqueue" relation.

If no search results found, allow to add given draft name as "not-received" relation (only if the draft exists in DT)

### DIFF
--- a/rpc/tests.py
+++ b/rpc/tests.py
@@ -161,7 +161,7 @@ class DocumentSearchTests(TestCase):
         )
 
         response = self.client.get(
-            "/api/rpc/documents/search/",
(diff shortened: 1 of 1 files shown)


jennifer-richards (changes requested):
Couple more race conditions (basically the same thing as in #1014), and we should start using `django.urls.reverse()` instead of hard-coding URLs in our tests.

PR #1014, opened 2026-03-19
feat: add draft to cluster on submission

rudimatz:
add logic to add docs to cluster on submission
   If any received reference is already clustered, the current document joins that
    cluster. Otherwise, create a new cluster and add the current document and all
    reference documents that are not already clustered.

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -10,7 +10,8 @@ import django_filters
 import rpcapi_client
 from django import forms
 from django.db import transaction
-from django.db.models import Coalesce, F, Max, Prefetch, Q, Subquery, Value
+from django.db.models import F, Max, Prefetch, Q, Subquery, Value
+from django.db.models.functions import Coalesce
 from django.http import JsonResponse
 from django.shortcuts import get_object_or_404
 from django.template.loader import render_to_string


rpc/api.py
```diff
@@ -150,6 +150,62 @@ def get_rfctobe_for_draft_name(draft_name: str) -> RfcToBe:
     return obj
 
 
+def _next_cluster_number() -> int:
+    return (Cluster.objects.aggregate(max_number=Max("number"))["max_number"] or 0) + 1
+
+
+def _add_doc_to_cluster(cluster: Cluster, doc: Document) -> None:
+    if ClusterMember.objects.filter(doc=doc).exists():
```
jennifer-richards: fyi, can also spell this `doc.clustermember_set.exists()`

rpc/api.py
```diff
@@ -150,6 +150,62 @@ def get_rfctobe_for_draft_name(draft_name: str) -> RfcToBe:
...
+
+    docs_by_id = {current_doc.id: current_doc}
+    for reference_doc in reference_docs:
+        docs_by_id.setdefault(reference_doc.id, reference_doc)
+
+    for doc in docs_by_id.values():
+        _add_doc_to_cluster(cluster, doc)
+
+    if not ClusterMember.objects.filter(cluster=cluster).exists():
+        cluster.delete()
```
jennifer-richards: Can you arrange to create the Cluster on demand rather than always creating, then deleting it? Not critical, but would be a bit cleaner.

rpc/api.py
```diff
@@ -486,19 +545,33 @@ def import_submission(request, document_id, rpcapi: rpcapi_client.PurpleApi):
...
                             "withdrawnref", rfctobe.pk, reference.name
                         )
                     else:
                         pass  # ignoring references to already published RfcToBe
 
+                    reference_doc = Document.objects.filter(
+                        datatracker_id=reference.id
+                    ).first()
+                    if reference_doc is not None:
+                        reference_docs.append(reference_doc)
```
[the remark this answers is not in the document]
rudimatz: I'll raise an exception. The Document is created at the beginning of function, so it should always be there

Issue #1036, opened 2026-03-27
Final Reviews table: Need to be able to remove approvers


ajeanmahoney:
Sometimes authors no longer want to be listed, or sometimes they are unavailable and their coauthors move them off the author list.

Issue #1031, opened 2026-03-27
final reviews: can't remove an action holder


ajeanmahoney:
It should be possible to remove an action holder from the Action Holders table.

PR #1056, opened 2026-04-01
feat: add buttons to delete actionholder and final reviews

rudimatz:
image

resolves https://github.com/ietf-tools/purple/issues/1036 and fixes https://github.com/ietf-tools/purple/issues/1031


rudimatz:
I added the buttons inside the "edit" dialog to guard them from accidental deletion

PR #1083, opened 2026-04-08
feat: add more complex history API

### LINKED ISSUE #873: Changing an editor assignment should be logged on the History tab
Currently an editor can be removed from an assignment, but that change isn't logged on the History tab, although the assignment does show as withdrawn on the Assignments tab.

### LINKED ISSUE #986: Changes to publishing dependencies should be logged
When I added a publishing dependency, it was saved in the publishing dependencies table, but it was not logged on the History tab.

### LINKED ISSUE #1126: note when documents become unblocked because their dependencies entered the queue
> Also note when documents become unblocked because their dependencies entered the queue. 

 _Originally posted by @ajeanmahoney in [#984](https://github.com/ietf-tools/purple/issues/984#issuecomment-4216898570)_

### LINKED ISSUE #1160: History tab: Stream Manager change lists a number rather than a person name
When I add a Stream Manager for a document (in this instance, I added Charles Eckel to https://purple.demo.rfc-editor.org/docs/draft-ietf-jmap-calendars, the History tab shows "Changed stream_manager from "None" to "348"". It should show "Changed stream_manager from "None" to "Charles Eckel".

### LINKED ISSUE #1165: History tab: Assigning a subseries number is not captured
When an RFC number is assigned, the History tab shows it (Changed rfc_number from "None" to "10001"). However, when a subseries number is assigned, there isn't a corresponding log entry on the History tab. This should also be logged.

### LINKED ISSUE #1170: History tab filter
It would be nice to be able to filter the rows on the History tab by subject, like "Show assignment updates" or "Show metadata changes"

rudimatz:
one history to rule them all:
merge histories of all relevant models for a RfcToBe into one

image

fix https://github.com/ietf-tools/purple/issues/1170 
fix https://github.com/ietf-tools/purple/issues/1165
fix https://github.com/ietf-tools/purple/issues/1160
fix https://github.com/ietf-tools/purple/issues/873
fix https://github.com/ietf-tools/purple/issues/986
fix https://github.com/ietf-tools/purple/issues/1126

### DIFF
--- a/client/app/pages/docs/[id]/history.vue
+++ b/client/app/pages/docs/[id]/history.vue
@@ -71,13 +71,13 @@ const currentTab: DocTabId = 'history'
 const draftName = computed(() => route.params.id?.toString() ?? '')
 
 type FilterValue = 'all' | 'assignment' | 'cluster' | 'reference' | 'author' | 'metadata'
-const FILTERS: { value: FilterValue; label: string; model: string | null }[] = [
-  { value: 'all', label: 'All', model: null },
-  { value: 'metadata', label: 'Metadata', model: 'rfctobe' },
-  { value: 'assignment', label: 'Assignment', model: 'assignment' },
-  { value: 'cluster', label: 'Cluster', model: 'cluster_member' },
-  { value: 'reference', label: 'Reference', model: 'reference' },
-  { value: 'author', label: 'Author', model: 'rfc_author' },
+const FILTERS: { value: FilterValue; label: string; models: string[] }[] = [
+  { value: 'all', label: 'All', models: [] },
+  { value: 'metadata', label: 'Metadata', models: ['rfctobe'] },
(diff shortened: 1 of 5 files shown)


client/app/pages/docs/[id]/history.vue
```diff
@@ -37,10 +46,10 @@
                     {{ entry.by.name }}
                   </NuxtLink>
                   <span v-else>
-                    {{ entry.by ? entry.by.name ? entry.by.name : '(nnnamed)' : '(System)' }}
```
Comment (jennifer-richards) [inline client/app/pages/docs/[id]/history.vue:40]: nnnice cccatch

jennifer-richards (approved):
Only consideration is that if there's no history, message indicating this might be helpful so the user can tell the UI is showing them an empty list (thinking "No events" or "No events of selected type" depending on whether a filter is applied). Just thinking out loud.

PR #1097, opened 2026-04-10
feat: display draft name in page title for tabs

### LINKED ISSUE #1037: Nit: display draftname in browser tab
The draftname is shown in the browser tab when the user is on the Info tab, but "RFC Production Center" is shown if the user is on any other tab for the document (Assignments, Approvals, Publication, History).

rudimatz:
resolves https://github.com/ietf-tools/purple/issues/1037

### DIFF
--- a/client/app/pages/clusters/[number].vue
+++ b/client/app/pages/clusters/[number].vue
@@ -40,7 +40,7 @@ definePageMeta({ validate: route => /^[1-9]\d*$/.test(route.params.number?.toStr
 
 const clusterNumber = computed(() => route.params.number ? parseInt(route.params.number.toString(), 10) : undefined)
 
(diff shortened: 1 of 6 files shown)


jennifer-richards (changes requested):
Minor, but I think `useHeadSafe` would be more appropriate. The draft name is well enough under our control that this isn't a serious security risk, but it shouldn't have an issue with being sanitized so why not err on the side of caution?

https://nuxt.com/docs/4.x/api/composables/use-head-safe

rudimatz:
@jennifer-richards ok updated to use useHeadSafe

rudimatz:
FYI @holloway

jennifer-richards (approved):

PR #1099, opened 2026-04-13
feat: only update RFC index on changes

jennifer-richards:
Adds the `DirtyBits` model as used in the errata app. Updates the `refresh_rfc_index_task()` to check whether the index is dirty before taking action. Marks the index as dirty when a new RFC is published, metadata changes to a published RFC are synced to datatracker, or changes are made to the unpublishable RFC set.

When deployed, we should schedule the task to run every 5-10 minutes.

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -1124,8 +1124,8 @@ class RfcToBeViewSet(viewsets.ModelViewSet):
             raise APIException(
                 f"Failed to sync metadata with datatracker: {err}"
             ) from err
-        else:
-            mark_rfcindex_as_dirty()
+
+        mark_rfcindex_as_dirty()
         return Response()
 
     @extend_schema(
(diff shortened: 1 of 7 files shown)


rpc/api.py
```diff
@@ -1123,7 +1124,8 @@ def sync_metadata(self, request, rpcapi: rpcapi_client.PurpleApi, draft__name=No
             raise APIException(
                 f"Failed to sync metadata with datatracker: {err}"
             ) from err
-
+        else:
+            mark_rfcindex_as_dirty()
         return Response()
```
rudimatz, line 1152: I think it would be slightly easier to read if the return was INSIDE the `else`

rudimatz (approved):

rjsparks (approved):

rpc/api.py
jennifer-richards: I'm going to take the `mark...()` call out of the `else` since it's making it less readable. Returning from within a `try...` construct will be sending people to the python docs unnecessarily. :-)

PR #1100, opened 2026-04-13
fix: not-received drafts shall not be added to a cluster

rudimatz:
don't create a ClusterMember for Normrefs that are not received


rudimatz:
relates to https://github.com/ietf-tools/rfced-xfer/pull/255

jennifer-richards (approved):

PR #1105, opened 2026-04-14
fix: cancel button on import

### LINKED ISSUE #996: nit: Cancel button doesn't work on Ingest Document page
But the user can use the browser's back arrow to back out of the page

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1086 and fix https://github.com/ietf-tools/purple/issues/996 and fix storing title on import

### DIFF
--- a/client/app/pages/docs/import.vue
+++ b/client/app/pages/docs/import.vue
@@ -120,7 +120,7 @@
     </div>
 
     <div class="mt-6 flex items-center justify-end gap-x-6">
-      <BaseButton btn-type="cancel">Cancel</BaseButton>
+      <BaseButton btn-type="cancel" @click="navigateTo('/queue/submissions')">Cancel</BaseButton>
       <BaseButton btn-type="default" :disabled="!haveRequiredValues" @click="importSubmission">Save</BaseButton>
     </div>
   </div>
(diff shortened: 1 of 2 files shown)


client/app/pages/docs/import.vue
```diff
@@ -120,7 +120,7 @@
     </div>
 
     <div class="mt-6 flex items-center justify-end gap-x-6">
-      <BaseButton btn-type="cancel">Cancel</BaseButton>
+      <BaseButton btn-type="cancel" @click="navigateTo('/queue/submissions')">Cancel</BaseButton>
```
holloway: Do we have a `const` from urls for this path?

should the path have a trailing slash?

rudimatz: the link via UI brings me to http://localhost:8088/queue/submissions
Do you think we should create a constant for this? It is basically just directing to the `submission` tab.

holloway: Although we're not doing it consistently, I think all URLs should have an entry in `urls.ts` rather than strings in lots of disparate files. The _link via UI_ will be another copy of that string so we should consolidate imo.

I guess the question of whether it should have a trailing slash is a broader question of Purple so I'm happy to ignore that for now.

rudimatz: I see, you mean like QUEUE_QUEUE_PATH?

holloway (approved):

PR #1107, opened 2026-04-14
fix: cluster reordering

rudimatz:
currently reordering fails for many clusters that contain PUBLISHED rfcs.

the published ones are filtered out by the API, they can be ignored for reordering, but they still require to have an order number since that field is NOT nullable.

Solution: make field "order" for ClusterMembers nullable, with a guard that will require it to be set for "active" drafts (created/in_progress)
When storing new order, set all INACTIVE draft's order to NULL. then only update the order field if it actually changed.

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -946,7 +946,7 @@ class ClusterViewSet(
             # Null out order for inactive members (published/withdrawn) so they don't
             # collide with the sequential order values assigned to active members.
             ClusterMember.objects.filter(cluster=cluster).exclude(
-                doc__rfctobe__disposition__slug__in=["created", "in_progress"]
+                doc__rfctobe__disposition__slug__in=DispositionName.ACTIVE_SLUGS
             ).update(order=None)
 
             for idx, draft_name in enumerate(draft_names, start=1):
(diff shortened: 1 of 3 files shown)


jennifer-richards (commented):
Looks ok to me in the details, but a couple high-level questions:

Will cluster / document histories show a log when the order is nulled for an inactive document?

I'm a little worried about having a nullable order field. I don't have a concrete issue to point to, but e.g., if we show a list of the documents in a cluster that _does_ include all of them, it should be in a stable order. Do we do that anywhere?

Another approach would be to keep the order as a non-nullable field and move inactive documents to the start (or end) of the range. E.g., when the first document in a cluster is published, it is moved to `order = 1` and the rest of the members are ordered starting at 2. When another is published, it becomes 2. This keeps `order` as a true ordering of the cluster, even though we're hiding some of the members, and maybe that's a feature

jennifer-richards (approved):
Discussed - I'm ok with nulling the order. We'll have simple history that we could use to reconstruct changes if we realize later that allowing null caused problems, and if an active doc accidentally becomes null somehow, it'll just show up at one end of the sort and a user will fix it when they notice.

PR #1108, opened 2026-04-14
feat: add API and UI for ClusterMember history

### LINKED ISSUE #984: Clusters: provide History tabs for clusters
On a cluster's page (for example, https://purple.staging.ietf.org/clusters/430/), provide a History tab that captures when documents are added and removed from the cluster

rudimatz:
resolves https://github.com/ietf-tools/purple/issues/984

- API endpoint for clustermember history
- add to API client schema
- display in UI a table with changes (add/remove/reorder)

### DIFF
--- a/client/app/components/ClusterMembershipHistory.vue
+++ b/client/app/components/ClusterMembershipHistory.vue
@@ -1,0 +1,55 @@
+<template>
+  <div>
+    <h2 class="text-lg font-semibold mb-3">Cluster-Members History</h2>
+    <div v-if="status === 'pending'" class="text-sm text-gray-500">Loading history...</div>
+    <div v-else-if="status === 'error'" class="text-sm text-red-500">Failed to load history.</div>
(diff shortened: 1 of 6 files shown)


jennifer-richards (changes requested):
This is fine to commit, but I don't think it fully resolves #984. There's still this [request to note unblocking events](https://github.com/ietf-tools/purple/issues/984#issuecomment-4216898570)

Also I'm worried that the reordering events will overwhelm the history with noise. The issue only asks for added/removed, so you _could_ drop them. If you keep them, I think it'd be better to try to group them into cluster-level events ("cluster reordered" rather than "document-a reordered; document-b reordered; document-d reordered, ..." every time someone adjusts a document). On this, I'm fine with taking it as it is and waiting to find out whether it's actually a problem.

jennifer-richards (approved):

rudimatz:
@jennifer-richards I made the mentioned comment into a separate issue https://github.com/ietf-tools/purple/issues/1126

PR #1114, opened 2026-04-16
feat: make std_level editable after publication

### LINKED ISSUE #1020: Document metadata: add current status
A document has publication status, that is, what it was at the time of publication (Proposed Standard, BCP, Standard, etc), and also a current status, which should match publication status until it is changed to Historic, Standard, etc. by the RPC at the request of the Stream.

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/1020 and show intended status


jennifer-richards (approved):
Ok. I wonder if the intended level should be labeled to make it clear that it is "draft was intended as...", though. Intended is a little ambiguous. 

(I also wonder if the intended level should be shown at all after the rfc is published, not sure whether that is relevant at that point)

PR #1117, opened 2026-04-16
feat: improve author search

### LINKED ISSUE #466: When adding authors on the author control, the author search hides results
Currently there is a limit on how many matching results are found, but no indication that there are more results that would be returned (hence different results shown if the user types more).

Instead, the control should either prompt the user to type more or have a way to see more results.

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/466

increase search result limit to 20 and show info message


rudimatz:
image

jennifer-richards (approved):

Issue #1116, opened 2026-04-16
Validate purple behavior if DT is unreachable


rudimatz:
Purple should behave graceful, but show error messages

rudimatz:
done with PR ^^

PR #1121, opened 2026-04-16
feat: handle DT unavailable

rudimatz:
relates to https://github.com/ietf-tools/purple/issues/1116
Add context manager for calls to DT API to have consistent behavior, throw errors if DT is not available and display as snackbar error in UI

Also add caching for RPC person profiles

### DIFF
--- a/datatracker/models.py
+++ b/datatracker/models.py
@@ -1,5 +1,6 @@
 # Copyright The IETF Trust 2023-2025, All Rights Reserved
 import rpcapi_client
+import urllib3.exceptions
 from django.core.cache import cache
 from django.db import models
 from simple_history.models import HistoricalRecords
@@ -82,6 +83,13 @@ class DatatrackerPerson(models.Model):
                 person = rpcapi.get_person_by_id(int(self.datatracker_id))
             except rpcapi_client.exceptions.NotFoundException:
                 cached_value = None
+            except (
+                urllib3.exceptions.MaxRetryError,
+                urllib3.exceptions.NewConnectionError,
+                rpcapi_client.exceptions.ApiException,
+            ):
+                # DT unavailable — return None without caching
+                return None
             else:
                 cached_value = person.json()
             cache.set(cache_key, cached_value)
@@ -159,6 +167,13 @@ class Document(models.Model):
                 document = rpcapi.get_draft_by_id(int(self.datatracker_id))
             except rpcapi_client.exceptions.NotFoundException:
                 cached_value = None
+            except (
+                urllib3.exceptions.MaxRetryError,
+                urllib3.exceptions.NewConnectionError,
+                rpcapi_client.exceptions.ApiException,
+            ):
+                # DT unavailable — return None without caching so next request retries
+                return None
             else:
                 cached_value = document.json()
             cache.set(cache_key, cached_value)
--- a/datatracker/models.py
+++ b/datatracker/models.py
@@ -5,7 +5,7 @@ from django.core.cache import cache
 from django.db import models
 from simple_history.models import HistoricalRecords
 
-from .rpcapi import with_rpcapi
+from .rpcapi import DataTrackerUnavailable, with_rpcapi
 from .utils import build_datatracker_url
 
 
@@ -87,9 +87,9 @@ class DatatrackerPerson(models.Model):
                 urllib3.exceptions.MaxRetryError,
                 urllib3.exceptions.NewConnectionError,
                 rpcapi_client.exceptions.ApiException,
-            ):
-                # DT unavailable — return None without caching
-                return None
+            ) as exc:
+                # DT unavailable — raise so callers can surface the error
+                raise DataTrackerUnavailable() from exc
             else:
                 cached_value = person.json()
             cache.set(cache_key, cached_value)
@@ -171,9 +171,8 @@ class Document(models.Model):
                 urllib3.exceptions.MaxRetryError,
                 urllib3.exceptions.NewConnectionError,
                 rpcapi_client.exceptions.ApiException,
-            ):
-                # DT unavailable — return None without caching so next request retries
-                return None
+            ) as exc:
+                raise DataTrackerUnavailable() from exc
             else:
                 cached_value = document.json()
             cache.set(cache_key, cached_value)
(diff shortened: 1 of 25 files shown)


datatracker/models.py
```diff
@@ -82,6 +83,13 @@ def _fetch(self, field_name, *, rpcapi: rpcapi_client.PurpleApi):
                 person = rpcapi.get_person_by_id(int(self.datatracker_id))
             except rpcapi_client.exceptions.NotFoundException:
                 cached_value = None
+            except (
+                urllib3.exceptions.MaxRetryError,
+                urllib3.exceptions.NewConnectionError,
+                rpcapi_client.exceptions.ApiException,
+            ):
+                # DT unavailable — return None without caching
+                return None
```
jennifer-richards: This worries me. If we do _anything_ with a return value that passes through  this that leaks out of purple, we risk corrupting data.

datatracker/models.py
```diff
@@ -159,6 +167,13 @@ def _fetch(self, field_name, *, rpcapi: rpcapi_client.PurpleApi):
                 document = rpcapi.get_draft_by_id(int(self.datatracker_id))
             except rpcapi_client.exceptions.NotFoundException:
                 cached_value = None
+            except (
+                urllib3.exceptions.MaxRetryError,
+                urllib3.exceptions.NewConnectionError,
+                rpcapi_client.exceptions.ApiException,
+            ):
+                # DT unavailable — return None without caching so next request retries
+                return None
```
jennifer-richards: As above, but even more worrisome here. I believe these are where values that land in new `RfcToBe` instances come from. A datatracker glitch will cause corrupt records.

rpc/api.py
```diff
@@ -319,24 +320,32 @@ class RpcPersonViewSet(viewsets.ReadOnlyModelViewSet, viewsets.GenericViewSet):
...
-        name_map = {
-            person.id: person.plain_name for person in rpcapi.get_persons(person_ids)
-        }
-        name_map |= {
-            missing_id: "Unknown"
-            for missing_id in person_ids
-            if missing_id not in name_map
-        }
+
+        name_map: dict[int, str] | None = cache.get(self._NAME_MAP_CACHE_KEY)
```
jennifer-richards: We could consider having a celery task that runs, say, every 10 minutes and refreshes the name map in memcache. If we set the cache lifetime to something longer than the task period, then views will never pay the penalty to refresh it, celery will always do the work.

rpc/tasks.py
```diff
@@ -284,7 +284,15 @@ def _compute_deep_references(
...
-        refs_2g = rpcapi.get_draft_references(target_dt_id) or []
+        try:
+            with datatracker_api():
+                refs_2g = rpcapi.get_draft_references(target_dt_id) or []
+        except Exception:
+            logger.warning(
+                "Datatracker unavailable fetching 2G refs for dt_id=%d; skipping",
+                target_dt_id,
+            )
+            continue
```
jennifer-richards: How do we recognize / recover from this happening? (Same question for the _3g code below)

rpc/signals.py
```diff
@@ -57,6 +59,14 @@ def final_approval_changed(sender, instance: FinalApproval, **kwargs):
     defer_apply(getattr(instance, "rfc_to_be", None))
 
 
+@receiver([post_save, post_delete], sender=RpcPerson)
+def rpc_person_changed(sender, instance: RpcPerson, **kwargs):
+    """Invalidate the person name map cache when RpcPerson records change."""
+    from .api import RpcPersonViewSet
+
+    cache.delete(RpcPersonViewSet._NAME_MAP_CACHE_KEY)
```
jennifer-richards: This seems like a broader cache invalidation than necessary. Only the datatracker person ID changing would require the map be updated. No need to refresh the cache on deletion, either - it's ok if the map has extra entries in it.

I think it might be worth some thought to whether there's a way to avoid a wholesale invalidation of the name map when a single reference is added. This could be adding fields to the value in cache (though this has possible race conditions), or doing explicit lookups of persons who aren't in the map as needed.

jennifer-richards (changes requested):
Unless I'm missing something, which is possible, I think there are a couple must fix issues:
* I'm very wary of returning `None` in place of datatracker values, I think that's sure to bite us badly with corrupt data. I'm not sure I see how to handle that except by failing and leaving it up to the calling code to decide how to proceed. In some places, sure, you can just say "<unknown>" and fail gracefully, but e.g. when importing a draft, if datatracker can't tell you its name, you just can't import it at all.
* Silently failing when computing references, relying on people watching/alerting on logs to notice something has gone wrong, is going to end up with stale data, isn't it? 

The name map caching will be an improvement, but I think it's still going to cause occasional delays in responses. See inline for some thoughts, though they're not fully baked. (In particular, the suggestion to use a celery task to keep the cache warm won't work without also sorting out how to deal with RpcPersons added between refreshes.)

rpc/tasks.py
rudimatz: @jennifer-richards we could add a "autoretry" here. Once DT is back, it would be processed

rudimatz:
latest changes: 
refresh cached RPC personnel whenever a person is not in current cache
retry deep refs when DT is unavailable

rpc/api.py
```diff
@@ -322,21 +325,23 @@ class RpcPersonViewSet(viewsets.ReadOnlyModelViewSet, viewsets.GenericViewSet):
...
+        if name_map is None or any(pid not in name_map for pid in person_ids):
+            with datatracker_api():
+                name_map = {
+                    person.id: person.plain_name
+                    for person in rpcapi.get_persons(person_ids)
+                }
+            cache.set(
+                RPC_PERSON_NAME_MAP_CACHE_KEY, name_map, RPC_PERSON_NAME_MAP_CACHE_TTL
+            )
+
```
jennifer-richards, line 352: If a deleted or invalid person ID ends up in purple, `name_map` may still not have it even after a fresh fetch. The return value from this method should have the `{missing_id: "Unknown"...}` patched back in to it, just don't cache those extra values. That will let purple continue to function.

jennifer-richards (changes requested):
I think one more change.

jennifer-richards (approved):

PR #1125, opened 2026-04-17
feat: RPC chair hold manager role

rudimatz:
give the manager role to RPC persons that hold the `chair` role for rpc in claim from DT

### DIFF
--- a/rpcauth/backends.py
+++ b/rpcauth/backends.py
@@ -7,10 +7,11 @@ import requests
 from django.core.exceptions import SuspiciousOperation
 from django.db import IntegrityError
 from django.utils.encoding import smart_str
-from rpc.models import RpcRole
 from josepy.jws import JWS, Header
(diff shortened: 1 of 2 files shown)


rpcauth/backends.py
```diff
@@ -173,8 +177,23 @@ def update_user(self, user, claims):
 
         if updated:
             user.save()
+        self._sync_manager_role(user, claims)
         return user
 
+    def _sync_manager_role(self, user, claims):
```
jennifer-richards, line 181: Small refactor request: rather than teaching this module about how we manage permissions, give that responsibility to `rpcauth.models.User`. If it had a method like `set_is_manager(value: bool)` to do the work of creating/removing the `can_hold_role` entry, then `backends.py` stays independent of the rpc app.

I think that's mostly just moving code, not actually changing it.

jennifer-richards (changes requested):

jennifer-richards (approved):

PR #1131, opened 2026-04-20
feat: Submit to crossref on publishing

kesara:


rpc/lifecycle/publication.py
```diff
@@ -394,16 +394,24 @@ def publish_rfctobe(
...
-        try:
-            submit_to_crossref(rfctobe.rpc_number)
-        except HTTPError as err:
-            raise PublicationError(
-                f"Successfully notified datatracker that RFC {rfctobe.rfc_number} "
-                f"was published, and files was uploaded. But crossref submission "
-                f"failed. HTTP Error: {err}"
-            ) from err
+        if all(
+            [settings.CROSSREF_API, settings.CROSSREF_USER, settings.CROSSREF_PASSWORD]
```
rjsparks: Please move these checks into the submit_to_crossref method itself. Consider making them explicit checks for None and even better, probe that they're in settings at all first.

rpc/lifecycle/publication.py
```diff
@@ -394,16 +394,24 @@ def publish_rfctobe(
...
+        ):
+            try:
+                submit_to_crossref(rfctobe.rpc_number)
+            except HTTPError as err:
+                raise PublicationError(
```
rjsparks: I think recovering from a, say transient. failure at CrossRef should is reasonable even _after_ we update Red - we've already updated datatracker. If we leave the calls in this order, I wouldn't raise PublicationError here.

If we're going to actually stop the publication process if crossref is misbehaving, we make the crossref post _before_ telling datatracker about anything.

rjsparks (dismissed):

rpc/lifecycle/publication.py
kesara: @jennifer-richards, What's the best way to report this error?

rpc/lifecycle/publication.py
kesara: I've made changes: https://github.com/ietf-tools/purple/pull/1131/changes/8507dee7cf0f060b6003a705788ac2d2667b0d1d

rpc/lifecycle/publication.py
jennifer-richards: Handling an error depends on how we want to notify the RPC about errors in notifying CrossRef. ~~Basic choices are to respond the way you've done (with modifications - see below) or to send email later.~~

Raising an unhandled `PublicationError` here is the way to say "don't do anything further in the task and don't raise a celery-level "task failed" exception. That goes for subclasses of `PublicationError` also. If it's a `TemporaryPublicationError` (or subclass of that), then it'll also be retried. Note that you cannot retry this task once datatracker accepts the RFC (i.e., where the task sets the disposition to "published") - it will fail if you republish after that.

Options:

1) ~~Do more or less what you've done here, except _after_ the `mark_rfcindex_as_dirty()` call. Before raising the `PublicationError`, call `record_failed_publication_attempt(rfctobe, msg)` where `msg` is the same message as your current exception. This will cause purple to display the message immediately to the RPC user who published the RFC.~~
2) Create a new task specifically for the CrossRef update set up to retry and, when retries are exhausted, email the admins. In the `publish_rfctobe_task()` after the existing code, schedule the new task with `kesaras_crossref_task.delay(rfctobe.id)`. This will cause purple to display a success message about publication and then, later, if the crossref task does not succeed before it runs out of retries, an email will be sent to alert people.

Sorry, I was mistaken. There's no mechanism to report a failed publication [through the purple UI] after the point-of-no-return. See next comment

jennifer-richards: Oops, got myself cross-eyed reading through the error handling. 

Important detail about unhandled `PublicationError` - it will immediately send the failure email, that's free.

The real option 1 is just:
1) Do what you've done here, but do it after `mark_rfcindex_as_dirty()` so that we update the index.

Option 2 will then give you retries before giving up and sending email.

jennifer-richards: (I think option 2 is worth the trouble, and if I understand @rjsparks correctly, he'd like those retries)

rpc/tasks.py
```diff
@@ -182,6 +184,9 @@ def publish_rfctobe_task(self, rfctobe_id, expected_head):
     rfctobe = RfcToBe.objects.get(pk=rfctobe_id)
     publish_rfctobe(rfctobe, expected_head=expected_head)
 
+    # Submit to corssref
```
jennifer-richards: typo: corssref -> crossref

rpc/tasks.py
```diff
@@ -347,3 +352,23 @@ def compute_deep_references_task(related_doc_id: int):
...
+
+
+class CrossrefSubmissionTask(RetryTask):
+    max_retries = 4 * 24 * 3  # every 15 minutes for 3 days
+
+
+@shared_task(
+    bind=True,
+    base=CrossrefSubmissionTask,
+    throws=(RfcToBe.DoesNotExist),
```
jennifer-richards: `throws` should be a tuple, so for a single entry that'd be `throws=(RfcToBe.DoesNotExist,),`

But you probably want `TemporaryPublicationError` (or its parent class, `PublicationError`) in the `throws` tuple so that celery won't make loud errors when it retries.

rpc/tasks.py
```diff
@@ -347,3 +352,23 @@ def compute_deep_references_task(related_doc_id: int):
...
+    autoretry_for=(TemporaryPublicationError,),
+)
+def crossref_submission_task(self, rfctobe_id):
+    rfc_number = RfcToBe.objects.get(pk=rfctobe_id).rfc_number
+    try:
+        submit_to_crossref(rfc_number)
+    except HTTPError as err:
+        raise TemporaryPublicationError(
```
jennifer-richards: This error handling belongs in the crossref.py module. Whenever possible, the task should be independent of the details of the methods it calls. 

Suggest that in crossref.py, you define something like `class CrossrefError(Exception)` and `class TemporaryCrossrefError(CrossrefError)` and raise those as appropriate for whether the task might reasonably succeed on retry.

That way, the task only knows that those exceptions are intentional in your code and that the latter exception means a retry could be attempted.

purple/crossref.py
```diff
@@ -175,22 +178,37 @@ def _generate_crossref_xml(rfc_number):
...
+            settings.CROSSREF_API,
+            data=data,
+            files=files,
+            timeout=settings.CROSSREF_TIMEOUT,
+        )
+        response.raise_for_status()
+    else:
+        logger.warning(
+            f"RFC {rfc_number} is not submitted crossref "
+            f"because crossref settings are not set."
```
jennifer-richards: Minor phrasing - suggest `"RFC {rfc_number} was not submitted to Crossref because Crossref settings are not configured."`

jennifer-richards (changes requested):
See inline - on the right track, but a bit more refactoring is needed

rpc/tasks.py
kesara: Fixed with 523fc849f20d7269d6d772529945ff72c23eb789

rpc/tasks.py
kesara: Addressed with f91b1c1ccd1fbc1d4e6f36f89edac31956097ba7

purple/crossref.py
kesara: Addressed with f91b1c1ccd1fbc1d4e6f36f89edac31956097ba7

rpc/tasks.py
kesara: I introduced a new class `CrossrefError` in f91b1c1ccd1fbc1d4e6f36f89edac31956097ba7
I didn't see a requirement for subclass.

jennifer-richards (changes requested):
Looks good to me. A test is failing - it looks like it's actually making HTTP requests to crossref. Need to get the test passing and make sure that it's not reaching out like that!

kesara:
> Looks good to me. A test is failing - it looks like it's actually making HTTP requests to crossref. Need to get the test passing and make sure that it's not reaching out like that!



> Looks good to me. A test is failing - it looks like it's actually making HTTP requests to crossref. Need to get the test passing and make sure that it's not reaching out like that!

No, the tests are mocking the responses.

jennifer-richards:
> 
> No, the tests are mocking the responses.

Oh good. The code didn't show up in the pr and the error message looked like it was live at a glance. Sorry about that.

jennifer-richards (approved):

PR #1142, opened 2026-04-22
fix: redo fetch metadata after fail, improve messages

### LINKED ISSUE #1046: Matadata validation should allow for re-fetch on error
Image

Currently, when requesting the metadata from the repo failed, pushing the button will only return the existing error over and over again.

Expected behavior: then the API returns an error (= a failed result) then the FE should first delete the existing result (using the existing function in client) and when try a new `create`

### LINKED ISSUE #1127: Fetch and verify metadata: error message uses incorrect RFC number
When a GitHub repo is not provided, and the user clicks the "Fetch and verify metadata" button on the Publication tab, the error message uses the incorrect RFC number. For example:

- draft-ietf-raw-architecture(RFC 9912): "Error: Details: No repository URL for RfcToBe 9784"
- draft-ietf-raw-technologies(RFC 9913): "Error: Details: No repository URL for RfcToBe 9775"
- draft-ietf-roll-dao-projection(RFC 9914): "Error: Details: No repository URL for RfcToBe 9770"

### LINKED ISSUE #1130: Fetch and verify metadata: purple.demo can't find GH repo
I've set up documents for publication, and have provided a valid repo for a document on the document's info page. When I click the "Fetch and verify metadata" button on Publication, I receive "Error: Details: No repository URL for RfcToBe 9784" (not the real RFC number. see #1127). When I click the Info tab and come back to the Publication tab, the message displayed is

Metadata validation failed: failed

Validation error

Metadata does not match and it cannot be published.

No git commit available in API response. Can't publish until this is verified.

No repository URL for RfcToBe 9784

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1130 and fix https://github.com/ietf-tools/purple/issues/1127 and fix https://github.com/ietf-tools/purple/issues/1046

general overhaul of the page:

- allow for deleting failed MetadataResults (without head_sha), which is required to re-fetch a previously failed request (which doesn't have the sha)
- show error message immediately if no repo set (don't call API)
- show dedicated error message if repo is set, but not reachable

Only show "Try again" if previous call failed, without the other data

### DIFF
--- a/client/app/pages/docs/[id]/publication-preparation.vue
+++ b/client/app/pages/docs/[id]/publication-preparation.vue
@@ -320,7 +320,7 @@ watch([rfcToBe, rfcToBeStatus, metadataValidationResultsStatus, publicationStatu
     step.value = {
       type: 'error',
       errorText: `Unable to load RFC ${draftName.value}. Server error: ${rfcToBeError.value?.message ?? ''} ${metadataValidationResultsError.value ?? ''}`,
-      showDeleteAndRetryButton: isMetadataValidationResults(precomputedResult) ? { headSha: precomputedResult.headSha ?? NO_HEAD_SHA_SENTINEL } : undefined,
+      showDeleteAndRetryButton: isMetadataValidationResults(precomputedResult) ? { headSha: precomputedResult.headSha! } : undefined,
       showResyncButton: true
     }
     return
@@ -429,7 +429,7 @@ const fetchAndVerifyMetadata = async () => {
     step.value = {
       type: 'error',
       errorText: `Couldn't start/poll for metadata sync results. Error: ${error}`,
-      showDeleteAndRetryButton: resultsCreate ? { headSha: resultsCreate.headSha ?? NO_HEAD_SHA_SENTINEL } : undefined,
+      showDeleteAndRetryButton: resultsCreate ? { headSha: resultsCreate.headSha } : undefined,
       showResyncButton: true
     }
     if (!resultsCreate) {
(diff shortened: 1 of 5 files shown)


client/app/pages/docs/[id]/publication-preparation.vue
```diff
@@ -466,29 +454,23 @@ const fetchAndVerifyMetadata = async () => {
...
-  if (!headSha) {
-    console.trace()
-    console.error("no git hash (head sha) found at step", step.value)
-    snackbar.add({ type: 'error', title: 'internal error, expected git hash (head sha) param', text: 'See dev console for more' })
-    return
-  }
+const deleteMetadataValidationAndRetry = async () => {
   step.value = { type: 'loading' }
   try {
     await api.documentsMetadataValidationResultsDestroy({
```
Comment (jennifer-richards) [inline client/app/pages/docs/[id]/publication-preparation.vue]: This delete API call will fail without the headSha parameter.

That was intended to prevent deleting a new result from a browser with stale data still loaded. I think the expected thing to do when a delete fails due to a headSha mismatch (i.e., 404) is to try to reload a new result with the latest headSha and show that, probably with a message something like "Your results were out of date, here are the current ones..."

(If there is no new one, then there's no error if a delete failed, just return to the starting state)

jennifer-richards (changes requested):
See inline - the `documentsMetadataValidationResultsDestroy()` backend will fail if it doesn't get a `headSha` parameter. I _think_ a different approach to failed delete will improve the behavior without having to get rid of that parameter, though.

rudimatz:
latest code reverts some previous changes. head_sha is required for delete, but a sentinel value may be passed to delete a null value

client/app/pages/docs/[id]/publication-preparation.vue
```diff
@@ -320,7 +320,7 @@ watch([rfcToBe, rfcToBeStatus, metadataValidationResultsStatus, publicationStatu
     step.value = {
       type: 'error',
       errorText: `Unable to load RFC ${draftName.value}. Server error: ${rfcToBeError.value?.message ?? ''} ${metadataValidationResultsError.value ?? ''}`,
-      showDeleteAndRetryButton: isMetadataValidationResults(precomputedResult) ? { headSha: precomputedResult.headSha ?? NO_HEAD_SHA_SENTINEL } : undefined,
+      showDeleteAndRetryButton: isMetadataValidationResults(precomputedResult) ? { headSha: precomputedResult.headSha! } : undefined,
```
Comment (jennifer-richards) [inline client/app/pages/docs/[id]/publication-preparation.vue:323]: We could probably tighten up the types so the explicit -! suffix is not needed, but it doesn't bother me.

jennifer-richards (approved):

PR #1153, opened 2026-04-24
feat: show publication validation/auth errors in UI

### LINKED ISSUE #1151: display error messages in publication flow
currently, the API shows errors why publication is blocked, but only in API response. They should be displayed in snackbar

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/1151

Will show errors/details why publication is blocked in UI/snackbar

I also changed the _gestalt_ of the JSON validation response (to streamline with Auth response):

```
Before: ["document has open assignments other than publisher"]

After: {"detail": "document has open assignments other than publisher"}
```

### DIFF
--- a/rpc/lifecycle/publication.py
+++ b/rpc/lifecycle/publication.py
@@ -60,7 +60,10 @@ def validate_ready_to_publish(rfctobe: RfcToBe):
     """
     if rfctobe.disposition_id != "in_progress":
         raise serializers.ValidationError(
-            {"disposition": f"disposition is '{rfctobe.disposition}', not 'In Progress'"},
+            {
+                "disposition": f"disposition is '{rfctobe.disposition}', "
+                "not 'In Progress'"
+            },
             code="rfctobe-bad-disposition",
         )
     if rfctobe.assignment_set.active().exclude(role_id="publisher").exists():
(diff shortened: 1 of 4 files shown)


rudimatz:
FYI @kesara

rpc/lifecycle/publication.py
```diff
@@ -60,36 +60,36 @@ def validate_ready_to_publish(rfctobe: RfcToBe):
     """
     if rfctobe.disposition_id != "in_progress":
         raise serializers.ValidationError(
-            f"disposition is '{rfctobe.disposition}, not 'In Progress'",
+            {"detail": f"disposition is '{rfctobe.disposition}, not 'In Progress'"},
```
jennifer-richards: This is a bit abusive of the `serializers.ValidationError` - see https://www.django-rest-framework.org/api-guide/exceptions/#validationerror. The keys of a dict parameter are meant to indicate which field has an error.

It'd probably be better to use a different `ApiException` class - I think that'll give you the `detail` key that you want.

jennifer-richards (changes requested):
See inline

jennifer-richards (approved):

PR #1173, opened 2026-04-28
fix: cluster handling for not-received

### LINKED ISSUE #1163: Cluster management: new document is not added to existing cluster
But unlike https://github.com/ietf-tools/purple/issues/997, the issue may be that the newly added document does not have any normref dependencies itself. 

To unblock Cluster 569 (https://purple.demo.rfc-editor.org/clusters/569/, which consists of draft-ietf-radext-reverse-coa and its missing reference draft-ietf-radext-radiusdtls-bis), I added draft-ietf-radext-radiusdtls-bis to the queue successfully, but it was not placed in C569. 

The info page for draft-ietf-radext-reverse-coa shows that its normative dependency draft-ietf-radext-radiusdtls-bis is in the queue, but C569's cluster page still shows that draft-ietf-radext-radiusdtls-bis has not been received.

### LINKED ISSUE #1164: Cluster management: New Cluster number not assigned
I added draft-ietf-httpapi-digest-fields-problem-types, which has a normative dependency on draft-ietf-httpbis-unencoded-digest, to the queue. On the Info tab for draft-ietf-httpapi-digest-fields-problem-types, draft-ietf-httpbis-unencoded-digest is listed as a normative dependency, but no cluster number is shown and no information about the document can be found on the Manage Clusters page. The addition of draft-ietf-httpapi-digest-fields-problem-types should have created Cluster 572.

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/1164 and fixes https://github.com/ietf-tools/purple/issues/1163

expected behavior:
when a new draft A enters the queue with a normative reference to draft B that is `not-received`, we should create a new cluster that contains (only) A.

When B eventually gets submitted, it will join A in same cluster.

### DIFF
--- a/rpc/tests.py
+++ b/rpc/tests.py
@@ -5,7 +5,6 @@ from datetime import date, timedelta
 from unittest.mock import MagicMock, patch
 
 import rpcapi_client
-
 from django.contrib.auth import get_user_model
 from django.core.cache import cache
 from django.test import TestCase
--- a/rpc/tests.py
+++ b/rpc/tests.py
@@ -1,23 +1,48 @@
 # Copyright The IETF Trust 2023, All Rights Reserved
 
 import json
-from unittest.mock import patch
+from datetime import date, timedelta
+from unittest.mock import MagicMock, patch
 
 from django.contrib.auth import get_user_model
+from django.core.cache import cache
 from django.test import TestCase
 from django.urls import reverse
 from rest_framework.exceptions import NotFound
 
-from rpc.models import DocRelationshipName
+from datatracker.factories import DocumentFactory
+from datatracker.models import Document
+from rpc.models import DocRelationshipName, RpcRelatedDocument
 
-from .api import resolve_rfctobe
+from .api import apply_submission_cluster_membership, resolve_rfctobe
 from .factories import (
     ClusterFactory,
     DispositionNameFactory,
     RfcToBeFactory,
+    SourceFormatNameFactory,
+    StdLevelNameFactory,
+    StreamNameFactory,
+    TlpBoilerplateChoiceNameFactory,
     UnusableRfcNumberFactory,
 )
 from .utils import next_rfc_number
+
+# Minimal data that rpcapi_client.FullDraft.from_json() accepts
+_FULL_DRAFT_BASE = {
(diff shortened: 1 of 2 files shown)


rpc/tests.py
```diff
@@ -149,6 +174,219 @@ def test_create_related_document_adds_target_to_existing_source_cluster(
...
+
+    def test_later_received_doc_joins_waiting_cluster(self):
+        """Draft B, imported after A, joins A's existing cluster."""
+        doc_a = RfcToBeFactory().draft
+        cluster_a = apply_submission_cluster_membership(
+            current_doc=doc_a,
+            reference_docs=[],
+            received_reference_ids=set(),
+            has_not_received_refs=True,
+        )
```
jennifer-richards, line 195: Just a nit-level thought about writing tests, not a call to change anything right now: it might be preferable to explicitly create a single doc cluster with `doc_a` in it rather than using `apply_submission_cluster_membership()` to do that work. This would eliminate the possibility that some untested side effect of the method is involved in making the test work. That leaves room for a bug if the cluster is created through some other mechanism.

It might be a place where a `ClusterFactory` would be useful in some future refactoring.

rpc/tests.py
```diff
@@ -149,6 +174,219 @@ def test_create_related_document_adds_target_to_existing_source_cluster(
...
+        """Return a mock rpcapi configured for the given drafts and references.
+
+        drafts_by_id: {doc_id: draft_name}
+        refs_by_id:   {doc_id: {ref_id: ref_name}} — defaults to no refs
+        """
+        refs_by_id = refs_by_id or {}
+        mock = MagicMock()
+
+        def _draft(doc_id, name):
+            d = MagicMock()
```
jennifer-richards: You might consider setting the `spec` on the Mock - if you passed in, say, `list(_FULL_DRAFT_BASE.keys()) + ['json']`, then attempts to access any properties not in that list would result in exceptions. (I think you could also pass it the class of the API type it's mocking and get a more accurate spec from that)

jennifer-richards (approved):
A couple of thoughts inline, but they're just thoughts/discussion about how we could write more effective tests.

PR #1176, opened 2026-04-28
feat: add logic for manual hold with comment

rudimatz:
image

BE:
- add new block reason `MANUAL HOLD` 
- this reason has some special logic attached compared to other block reasons:
-- it is added by API call and can only be removed by API call (not signals)
- add comment for block reason model

UI:
- add button to set and clear a manual hold
- setting it will open a modal to set an optional comment for the manual hold
- improve ordering of assignments (incl block assignments) to reflect the chronological order
- show block reasons and comments


rpc/lifecycle/blocked_assignments.py
```diff
@@ -288,10 +289,11 @@ def _close_blocked_assignments(rfc: RfcToBe) -> bool:
...
 
-    # Resolve all active blocking reasons
-    RfcToBeBlockingReason.objects.filter(rfc_to_be=rfc, resolved__isnull=True).update(
-        resolved=timezone.now()
-    )
+    # Resolve all active blocking reasons except manual_hold, which only the
+    # explicit API action may clear.
+    RfcToBeBlockingReason.objects.filter(rfc_to_be=rfc, resolved__isnull=True).exclude(
+        reason__slug=BlockingReason.MANUAL_HOLD
+    ).update(resolved=timezone.now())
```
jennifer-richards: Do we need to make an explicit `save()` call to preserve history?

rpc/lifecycle/blocked_assignments.py
```diff
@@ -336,3 +338,66 @@ def apply_blocked_assignment_for_rfc(rfc: RfcToBe) -> bool:
...
+    """Store a manual hold reason and create a blocked assignment if needed.
+
+    If the RFC is not yet blocked: closes active assignments to CLOSED_FOR_HOLD
+    and creates a new blocked assignment. If already blocked (e.g. an automatic
+    block is active), only the reason record is added — no assignment changes.
+    """
+    try:
+        with transaction.atomic():
+            locked = RfcToBe.objects.select_for_update().get(pk=rfc.pk)
+            reason = BlockingReason.objects.get(slug=BlockingReason.MANUAL_HOLD)
```
jennifer-richards: It may eliminate a query if you cut out the `BlockingReason.objects.get()` and instead create the record with `...create(rfc_to_be=locked, reason_id=BlockingReason.MANUAL_HOLD, comment=comment)`. Sometimes the ORM is clever enough to figure this out, but not always. This is used a lot in datatracker. Not essential here, though.

rpc/lifecycle/blocked_assignments.py
```diff
@@ -336,3 +338,66 @@ def apply_blocked_assignment_for_rfc(rfc: RfcToBe) -> bool:
...
+            RfcToBeBlockingReason.objects.create(
+                rfc_to_be=locked, reason=reason, comment=comment
+            )
+            if _has_active_blocked_assignment(locked):
+                logger.info(
+                    "RFC %s already blocked; skip assignment creation for manual hold",
+                    locked.pk,
+                )
+                return
+            _create_blocked_assignments(locked, reasons={BlockingReason.MANUAL_HOLD})
```
jennifer-richards: I'm a little puzzled. When `_has_active_blocked_assignment(locked)` is `False`, I think you create the `RfcToBeBlockingReason` and then immediately call `_create_blocked_assignments()` which tries to create it again (but you've modified it to use `get_or_create()`). Is there a reason this object needs to possibly be created in two different places? It feels a bit like spaghetti code beginning to form and I wonder if there's a (reasonable) refactor that could avoid that.

jennifer-richards (changes requested):
Calling this "Request changes," but it's really "Request discussion" right now. See inline, particularly the question about creating `RfcToBeBlockingReason` instances in multiple spots.

rpc/lifecycle/blocked_assignments.py
rudimatz: yes indeed. I didn't think about it here, but good point

rudimatz:
all 3 fixed

jennifer-richards (approved):

PR #1180, opened 2026-04-29
feat: let history be written for stream manager

### LINKED ISSUE #1155: History tab: Status change msg unclear, impacts publication date?
When I change an RFC's status from, say, STD to Historic, the document's History tab shows the following:

`	Changed published_at from "None" to "2022-06-06 20:08:47+00:00"; Changed std_level from "std" to "hist"`

Why does `published_at` show a change? The document (RFC 9110 in this case) was published on 2022-06-06. Changing the document's status should not impact publication date.  Is the timestamp (20:08:47+00:00) from the production rfced database?

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1155

current implementation skips history and makes the change appear in history randomly with the next `save()`

### DIFF
--- a/rpc/tasks.py
+++ b/rpc/tasks.py
@@ -45,7 +45,8 @@ def set_stream_manager_task(rfc_to_be_id: int):
     if rfctobe.stream_manager_id is not None:
         return
     person = rfctobe.resolve_stream_manager_person()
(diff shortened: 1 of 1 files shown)


jennifer-richards (approved):

rudimatz:
related task https://github.com/ietf-tools/rfced-xfer/pull/262

PR #1183, opened 2026-04-30
fix: make consistent behavior resolve draft identifier

### LINKED ISSUE #1154: Info page: HTTP 500 returned for comments on RFC page, but not for its I-D page
When comparing the contents of the Info page for an RFC URL (e.g., [https://purple.staging.rfc-editor.org/docs/rfc 3261](https://purple.staging.rfc-editor.org/docs/rfc%203261) and its equivalent draftstring URL (https://purple.staging.rfc-editor.org/docs/draft-ietf-sip-rfc2543bis) (draft-ietf-sip-rfc2543bis became RFC 3261), the RFC URL has the following error for the comments section: "Problem loading comments: (HTTP 500) Response returned an error code". The draftstring URL does not.

Another example: https://purple.staging.rfc-editor.org/docs/draft-ietf-6man-rfc2460bis and [https://purple.staging.rfc-editor.org/docs/rfc 8200](https://purple.staging.rfc-editor.org/docs/rfc%208200)

### LINKED ISSUE #1161: nit: Ability to search for "rfc nnnn" in addition to "rfcnnnn"
When "rfc 2119" is typed in the search field, no matches are found. When "rfc2119" is typed, the info pages for both RFC 2119 and RFC 8174 are found (which is correct because RFC 8174 was draft-leiba-rfc2119-update).

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/1154 and fixes https://github.com/ietf-tools/purple/issues/1161

make behavior consistent to find RfcToBe by identifier
now, both "rfc 1234" and "rfc1234" work, both in URL and via search
use existing resolver to avoid DRY

If error, show API response in error message in UI

### DIFF
--- a/client/app/pages/docs/[id]/index.vue
+++ b/client/app/pages/docs/[id]/index.vue
@@ -6,7 +6,7 @@
 
     <div class="mx-auto max-w-7xl px-4 py-8 sm:px-6 lg:px-8">
       <ErrorAlert v-if="rawRfcToBeError" title="API Error">
-        API error while requesting draft: {{ rawRfcToBeError }}
+        {{ rfcToBeErrorMessage }}
       </ErrorAlert>
       <div v-else
         class="mx-auto grid max-w-2xl grid-cols-1 grid-rows-1 place-items-stretch gap-x-8 gap-y-8 lg:mx-0 lg:max-w-none lg:grid-cols-3">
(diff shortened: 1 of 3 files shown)


jennifer-richards (approved):
Looks ok, approved...

But a high level question: would it be better not to generate URLs with the space and avoid having to do this munging? It's not difficult, but it creates some maintenance burden to keep supporting it.

rudimatz:
@jennifer-richards yes I shared your concern. But I think here here it makes sense, in the special case of "RFCs without a draft", the string `RFC 123` is the string identifier internally, and the behavior overall is consistent

PR #1186, opened 2026-04-30
feat: add logic to start assignment

### LINKED ISSUE #1150: add logic to start an assignment
after a RPC members gets an assignment, the state is "assigned". There needs to be a way to "START" is, and change the status to "in_progress"

proposal:
rename "Finish assignments" to "Manage Assignments"
and rename button "Finish" to "Start", if that current assignment is not in_progress

rudimatz:
resolve https://github.com/ietf-tools/purple/issues/1150


rudimatz:
use a utils function for sorting Assignments in consistent way

jennifer-richards (approved):

PR #1188, opened 2026-05-01
feat: move cluster in same of already received ref

rudimatz:
improve logic:
if a new draft is imported and either it is or it has a normative reference already in cluster, it should join the same cluster.
if either it is or it has a normative reference that is NOT yet in a cluster, the cluster shall be created and both should join the cluster

- display error in enqueuing in snackbar
- in cluster graph, show disposition only if not blocked
- add tests

### DIFF
--- a/rpc/utils.py
+++ b/rpc/utils.py
@@ -60,6 +60,7 @@ def create_cluster() -> Cluster:
         Value(1),
     )
     cluster = Cluster.objects.create(number=next_number)
+    # Refresh from DB to get the actual assigned number, not Coalesce expression
     cluster.refresh_from_db()
     return cluster
 
--- a/rpc/utils.py
+++ b/rpc/utils.py
@@ -59,7 +59,9 @@ def create_cluster() -> Cluster:
         ),
         Value(1),
     )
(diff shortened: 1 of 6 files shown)


rpc/utils.py
```diff
@@ -59,7 +59,9 @@ def create_cluster() -> Cluster:
         ),
         Value(1),
     )
-    return Cluster.objects.create(number=next_number)
+    cluster = Cluster.objects.create(number=next_number)
+    cluster.refresh_from_db()
```
jennifer-richards, line 64: Why the refresh? I _think_ that the `create()` should be reading back the values from the `INSERT` and updating the model accordingly. Does it not do that?

jennifer-richards (approved):
question inline, but approved

rpc/utils.py
rudimatz: yes it does, but it was not doing it in a clean way for the "next_number", which was returned as a `Coalesce` object, not as a `number` as expected

rudimatz: adding explaining comment

PR #1200, opened 2026-05-05
fix: store datepicker dates as noon, display as UTC

rudimatz:
apply changes for fields
- external deadline (Rfc)
- approved / requested (Final Approval)
- deadline / completed (Action holder)

### DIFF
--- a/client/app/utils/form.ts
+++ b/client/app/utils/form.ts
@@ -26,5 +26,6 @@ export const hoursToDurationString = (hours: number): string => {
 
 export const jsDateToInputTypeDate = (date: Date): string => DateTime.fromJSDate(date, { zone: 'utc' }).toISODate() ?? ''
 
+// use 12:00 UTC to minimize date ambiguity across time zones
 export const inputTypeDateToDateTime = (isoDateString: string): DateTime => DateTime.fromISO(isoDateString, { zone: 'utc' }).set({ hour: 12 })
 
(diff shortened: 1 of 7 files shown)


client/app/utils/form.ts
```diff
@@ -24,7 +24,7 @@ export const hoursToDurationString = (hours: number): string => {
   return duration.toFormat('d hh:mm:ss')
 }
 
-export const jsDateToInputTypeDate = (date: Date): string => DateTime.fromJSDate(date).toISODate() ?? ''
+export const jsDateToInputTypeDate = (date: Date): string => DateTime.fromJSDate(date, { zone: 'utc' }).toISODate() ?? ''
 
-export const inputTypeDateToDateTime = (isoDateString: string): DateTime => DateTime.fromISO(isoDateString, { zone: 'utc' })
+export const inputTypeDateToDateTime = (isoDateString: string): DateTime => DateTime.fromISO(isoDateString, { zone: 'utc' }).set({ hour: 12 })
```
jennifer-richards, line 30: A comment explaining the choice of `hour: 12` would be helpful to future developers. Suggest something like `use 12:00 UTC to minimize date ambiguity across time zones`

jennifer-richards (commented):
While we're thinking about it, I think it'd be worthwhile to comment on the datetime fields in the model that are actually treated as dates by the UI. If that's a large ask, push back, but at least a code comment would be nice.

rudimatz:
agreed upon solution: for date fields populated by user input (datepicker), store them as noon UTC.
When displayed in UI, show the UTC date (which will then always match the date selected as a `string` ) - regardless of user;s locale.

Other dates that are set by the backend (e.g. `publishedAt`) continue to be displayed in user's local time

jennifer-richards (approved):

PR #1202, opened 2026-05-05
feat: add rfc number in pub queue payload

rudimatz:


rudimatz:
required by https://github.com/ietf-tools/queue/issues/30

jennifer-richards (approved):

Issue #869, opened 2026-02-11
Should not be able to assign subseries number unless status matches


ajeanmahoney:
The user can set a subseries number even if the status (that is, the "Standard Level") doesn't match. 
There should be a check against this

rudimatz:
@alicerusso
can you explain how exactly those 2 fields shall interact.
What should be the "status" field if its a FYI subseries?

<img width="144" height="147" alt="Image" src="https://github.com/user-attachments/assets/797d5ef5-3c12-4c4e-9beb-4c09d5a33130" />

alicerusso:
an FYI is Informational. 
an STD is Internet Standard.
a BCP is Best Current Practice.

PR #1203, opened 2026-05-06
fix: improve label for level and search placeholder

rudimatz:
label name change as suggested in https://github.com/ietf-tools/purple/issues/869


client/app/components/HeaderNav.vue
```diff
@@ -13,7 +13,7 @@
         <label for="search-field" class="sr-only">Search</label>
         <Icon name="uil:search" class="pointer-events-none absolute inset-y-0 left-0 h-full w-5 text-gray-400" aria-hidden="true" />
         <input id="search-field" v-model="navSearch" class="block h-full w-full border-0 py-0 pl-8 pr-0 bg-white dark:bg-neutral-900 text-gray-900 dark:text-neutral-200 placeholder:text-gray-400 dark:placeholder:text-neutral-400 focus:ring-0 sm:text-sm"
-          placeholder="Search docs by draft name, RFC number or author" type="search" name="search" autocomplete="off" @focus="showResults = true" @blur="onSearchBlur" >
+          placeholder="Search docs by draft name, RFC number, Subseries (e.g. BCP5) or author" type="search" name="search" autocomplete="off" @focus="showResults = true" @blur="onSearchBlur" >
```
jennifer-richards: Should there be a space in the example ("BCP 5")?

jennifer-richards (approved):
Question inline, but ok if it's already as it should be

Issue #590, opened 2025-11-08
/queue/published loads slowly


rjsparks:
* Speed up the load
* Add a pacifier until load is quick enough that it's not needed.

Issue #222, opened 2024-09-30
Detecting N+1 queries


microamp:
Detecting N+1 queries early in the development process would be beneficial. 

Scout: https://scoutapm.com/blog/django-and-the-n1-queries-problem

Additionally, other tools are available for use during local development.

- https://github.com/jmcarp/nplusone
- https://github.com/Tijani-Dia/dj-tracker

Let me know if we are already doing something similar internally.

microamp:
Example: `/api/rpc/rpc_person/`

It first runs

``` python
RpcPerson.objects.all()
```

to get all instances of `RpcPerson`, which runs the following SQL query under the hood.

``` sql
SELECT "rpc_rpcperson"."id", 
       "rpc_rpcperson"."datatracker_person_id",
       "rpc_rpcperson"."hours_per_week", 
       "rpc_rpcperson"."manager_id"
FROM "rpc_rpcperson"
```

And then, for each row in the result set, it will spawn two additional queries (i.e. 2N+1); one for capabilities, another for roles. 

Since there will be only a small number of users, I don't think it will be a problem performance-wise - it's just an example.

rjsparks:
Much of the infrastructure so far is designed to drive the iterative review process with the RPC - it's just enough to get UI working just enough to get feedback. 

We are close, but not quite to, the point that we have a stable enough idea of what the queries need to return that we should focus on performance. We know, for instance, that most of the early api endpoints dramatically over-return for the views needs, and they will need to be significantly refactored before we optimize them.

rudimatz:
Started profiling to find the worst/slowest queries

Improved Clustermember query in https://github.com/ietf-tools/purple/pull/530

PR #1207, opened 2026-05-06
feat: add cache warming for DT persons

rudimatz:
current behavior: we call DT person names for each DT person that comes up in a API response.
We do cache those... but since we populate usually all of them at the same time, they will be invalidated from cache all at the same time and N calls have to be done to refresh the cache.

Better approach: use a batch endpoint, and populate cache before the API response is processed.


rudimatz:
improves https://github.com/ietf-tools/purple/issues/222 and fixes https://github.com/ietf-tools/purple/issues/590

jennifer-richards (approved):

PR #1210, opened 2026-05-06
fix: error adding action holder

### LINKED ISSUE #1209: error adding Action holder
reported by @sginoza
getting 500 Error when adding "IAB Chair" as action holder

Image

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1209

create action holder failed if the dt person was not yet created (if it was not by chance already an RFC author or otherwise imported)


rudimatz:
also nit: fix field label

jennifer-richards (approved):

PR #1216, opened 2026-05-07
fix: vitest error fix and run tests for every PR towards main

rudimatz:
relates to https://github.com/ietf-tools/purple/pull/1215

I also changed the logic to run these tests for every PR towards main. Is that fine with you @holloway ?


holloway (approved):

holloway:
@rudimatz absolutely!

PR #1222, opened 2026-05-07
feat: add not-received in list of cluster docs

### LINKED ISSUE #58: Clusters page: List not-received docs
Should look like https://purple.staging.rfc-editor.org/clusters

rudimatz:
this changes should fix https://github.com/ietf-tools/queue/issues/58 and make queue cluster page work as it is.

Reasoning: `not-received` docs are not stored as ClusterMembers. However they should appear like cluster members on the queue site.
Solution: custom PublicClusterMemberSerializer that will add those `not-received` on-the-fly to the list as if they were full ClusterMembers

### DIFF
--- a/purple_api.yaml
+++ b/purple_api.yaml
@@ -18,7 +18,7 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: '#/components/schemas/Cluster'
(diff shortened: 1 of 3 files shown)


rudimatz:
OBE by https://github.com/ietf-tools/queue/pull/68

PR #1227, opened 2026-05-08
fix: display errors on pub and save attempt

### LINKED ISSUE #1224: MissingFilesError during publish_rfctobe_task does not mark PublicationAttempt as failed
See title. If the repo / manifest is missing files required for publication, the `choose_files()` method raises a `MissingFilesError`. This should be caught and result in the `PublicationAttempt` being marked as failed with the descriptive error message in the `detail` field. It should _not_ bubble up as an unhandled exception that triggers an error email to be sent.

While at it, should make sure that _any_ exception anywhere in that method sets the PublicationAttempt to failed. For these the email _should_ be sent, so the error should be re-raised.

Backtrace from the error
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/celery/app/trace.py", line 585, in trace_task
    R = retval = fun(*args, **kwargs)
                 ~~~^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/celery/app/trace.py", line 858, in __protected_call__
    return self.run(*args, **kwargs)
           ~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/celery/app/autoretry.py", line 38, in run
    return task._orig_run(*args, **kwargs)
           ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/workspace/rpc/tasks.py", line 189, in publish_rfctobe_task
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/datatracker/rpcapi.py", line 70, in wrapper
    return f(*args, **kwargs)
  File "/workspace/rpc/lifecycle/publication.py", line 318, in publish_rfctobe
    chosen_files = choose_files(publication["files"])
  File "/workspace/rpc/lifecycle/publication.py", line 109, in choose_files
    raise MissingFilesError(f"Missing files: {', '.join(missing)}")
rpc.lifecycle.publication.MissingFilesError: Missing files: txt, html
```

rudimatz:
also: show errors in snackbar, add tests

fixes https://github.com/ietf-tools/purple/issues/1224

### DIFF
--- a/rpc/lifecycle/publication.py
+++ b/rpc/lifecycle/publication.py
@@ -246,12 +246,12 @@ def record_failed_publication_attempt(rfctobe: RfcToBe, detail: str):
                     f"PublicationAttempt created by another process during execution "
                     f"of record_failed_publication_attempt() for {rfctobe}"
                 )
-        elif pub_attempt.status != PublicationAttempt.Status.PENDING:
-            logger.warning(
-                f"PublicationAttempt status was {pub_attempt.status} when recording "
-                f"failure for {rfctobe}"
-            )
         else:
+            if pub_attempt.status != PublicationAttempt.Status.PENDING:
+                logger.warning(
+                    f"PublicationAttempt status was {pub_attempt.status} when "
+                    f"recording failure for {rfctobe}"
+                )
             pub_attempt.status = PublicationAttempt.Status.FAILED
             pub_attempt.detail = detail
             pub_attempt.save()
@@ -271,14 +271,32 @@ def clear_failed_publication_attempt(rfctobe: RfcToBe):
 def publish_rfctobe(
     rfctobe: RfcToBe, expected_head: str, *, rpcapi: rpcapi_client.PurpleApi
 ):
+    try:
+        _do_publish_rfctobe(rfctobe, expected_head=expected_head, rpcapi=rpcapi)
+    except PublicationError:
+        raise
+    except Exception:
+        record_failed_publication_attempt(
+            rfctobe, "Unexpected error during publication"
+        )
+        raise
+
+
+def _do_publish_rfctobe(
+    rfctobe: RfcToBe, expected_head: str, *, rpcapi: rpcapi_client.PurpleApi
+):
     # Re-validate that the RfcToBe is ready to publish, things may have changed
     # since the task was queued.
     try:
         validate_ready_to_publish(rfctobe)
     except serializers.ValidationError as err:
         first = next(iter(err.detail.values()), None)
-        msg = str(first) if first is not None else str(err.detail)
-        raise PublicationError(f"Cannot publish because {msg}") from err
+        msg = (
+            f"Cannot publish because "
+            f"{str(first) if first is not None else str(err.detail)}"
+        )
+        record_failed_publication_attempt(rfctobe, msg)
+        raise PublicationError(msg) from err
 
     repo = GithubRepository(rfctobe.repository)
     # Check that head commit matches expected_head. This check defines the instant
@@ -315,7 +333,11 @@ def publish_rfctobe(
         record_failed_publication_attempt(rfctobe, msg)
         raise PublicationError(msg)
     # Choose files + validate that we have what we need / no ambiguities
-    chosen_files = choose_files(publication["files"])
+    try:
+        chosen_files = choose_files(publication["files"])
+    except (MissingFilesError, AmbiguousFilesError) as err:
+        record_failed_publication_attempt(rfctobe, str(err))
+        raise
     downloaded_files = {}
     # Download the selected files to a temp directory
     with TemporaryDirectory(delete=not settings.DEBUG) as tmpdirname:
@@ -389,11 +411,13 @@ def publish_rfctobe(
             # The RFC was already published, but the files were not accepted.
             # This situation requires manual intervention until we make the
             # publication notification idempotent.
-            raise PublicationError(
+            detail = (
                 f"Successfully notified datatracker that RFC {rfctobe.rfc_number} "
                 f"was published, but uploading its files failed. Manual correction "
                 f"is required."
-            ) from err
+            )
+            record_failed_publication_attempt(rfctobe, detail)
+            raise PublicationError(detail) from err
         else:
             mark_rfcindex_as_dirty()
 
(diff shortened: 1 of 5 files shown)


rpc/lifecycle/publication.py
```diff
@@ -270,15 +270,33 @@ def clear_failed_publication_attempt(rfctobe: RfcToBe):
...
     except serializers.ValidationError as err:
         first = next(iter(err.detail.values()), None)
-        msg = str(first) if first is not None else str(err.detail)
-        raise PublicationError(f"Cannot publish because {msg}") from err
+        msg = (
+            f"Cannot publish because "
+            f"{str(first) if first is not None else str(err.detail)}"
+        )
+        record_failed_publication_attempt(rfctobe, msg)
+        raise PublicationError(msg) from err
```
jennifer-richards: I think this should be a `return` rather than a `raise` - the _task_ completed successfully. The _publication_ failed, but refusing to publish and ending the attempt with an error when something goes wrong is part of the task's job.

rpc/lifecycle/publication.py
```diff
@@ -315,7 +333,11 @@ def publish_rfctobe(
         record_failed_publication_attempt(rfctobe, msg)
         raise PublicationError(msg)
     # Choose files + validate that we have what we need / no ambiguities
-    chosen_files = choose_files(publication["files"])
+    try:
+        chosen_files = choose_files(publication["files"])
+    except (MissingFilesError, AmbiguousFilesError) as err:
+        record_failed_publication_attempt(rfctobe, str(err))
+        raise
```
jennifer-richards: Again, `return` not `raise`

rpc/lifecycle/publication.py
```diff
@@ -389,11 +411,13 @@ def publish_rfctobe(
...
             # publication notification idempotent.
-            raise PublicationError(
+            detail = (
                 f"Successfully notified datatracker that RFC {rfctobe.rfc_number} "
                 f"was published, but uploading its files failed. Manual correction "
                 f"is required."
-            ) from err
+            )
+            record_failed_publication_attempt(rfctobe, detail)
+            raise PublicationError(detail) from err
```
jennifer-richards: This time, yes, raise because this will require admin intervention, we want to alert admins directly. I think just raising the `PublicationError` will cause the task's handler to record the failed attempt. It'd be best not to call it twice but should work if you do.

rpc/lifecycle/publication.py
```diff
@@ -246,12 +246,12 @@ def record_failed_publication_attempt(rfctobe: RfcToBe, detail: str):
                     f"PublicationAttempt created by another process during execution "
                     f"of record_failed_publication_attempt() for {rfctobe}"
                 )
-        elif pub_attempt.status != PublicationAttempt.Status.PENDING:
-            logger.warning(
-                f"PublicationAttempt status was {pub_attempt.status} when recording "
-                f"failure for {rfctobe}"
-            )
         else:
+            if pub_attempt.status != PublicationAttempt.Status.PENDING:
```
jennifer-richards, line 250: I think it'd be best not to change `pub_attempt.status` unless it was `PENDING` when you found it. It's ok to silence the warning if it's already in the `FAILED` state, but it'd be best not to overwrite the detail. (I am guessing you're doing this to eliminate warnings caused by calling this method twice - see other comments about avoiding that)

(This logic all needs some attention now that we have a better handle on how the publication process works. I think the existing branch that creates a `PublicationAttempt` should also not be there, because it'll risk creating a failed attempt record for an `RfcToBe` that was already successfully published. I believe that's harmless, but it creates cruft. That can wait for later)

jennifer-richards (changes requested):
Looks great, but I think a couple tweaks are needed. See inline.

rpc/lifecycle/publication.py
```diff
@@ -303,19 +316,22 @@ def publish_rfctobe(
     except TemporaryRepositoryError as err:
         raise TemporaryPublicationError("Error retrieving manifest") from err
     except RepositoryError as err:
-        msg = "Invalid or missing manifest"
-        record_failed_publication_attempt(rfctobe, msg)
-        raise PublicationError(msg) from err
+        raise PublicationError("Invalid or missing manifest") from err
```
jennifer-richards: This is probably another one that should not raise an error - a missing manifest is something the person trying to publish should be able to fix. It doesn't need admins / email backtrace.

rpc/lifecycle/publication.py
```diff
@@ -270,27 +270,40 @@ def clear_failed_publication_attempt(rfctobe: RfcToBe):
 @with_rpcapi
 def publish_rfctobe(
     rfctobe: RfcToBe, expected_head: str, *, rpcapi: rpcapi_client.PurpleApi
+):
+    _do_publish_rfctobe(rfctobe, expected_head=expected_head, rpcapi=rpcapi)
+    # Submit to crossref only if publication succeeded
+    if rfctobe.disposition_id == "published":
+        from rpc.tasks import crossref_submission_task
+
+        crossref_submission_task.delay(rfctobe_id=rfctobe.id)
```
jennifer-richards, line 282: Ideally `crossref_submission_task` should only be queued if `rfctobe.disposition_id` changed from `in_progress` to `published`. Easiest would be to check the disposition before calling `_do_publish_rfctobe()` - it's not perfect but is straightforward.

rpc/lifecycle/publication.py
```diff
@@ -246,12 +246,12 @@ def record_failed_publication_attempt(rfctobe: RfcToBe, detail: str):
...
-            logger.warning(
-                f"PublicationAttempt status was {pub_attempt.status} when recording "
-                f"failure for {rfctobe}"
-            )
         else:
+            if pub_attempt.status != PublicationAttempt.Status.PENDING:
+                logger.warning(
+                    f"PublicationAttempt status was {pub_attempt.status} when "
+                    f"recording failure for {rfctobe}"
+                )
```
jennifer-richards, line 254:
```suggestion
                    f"PublicationAttempt status was {pub_attempt.status} when "
                    f"recording failure for {rfctobe}. Detail was: {pub_attempt.detail}"
                )
```

jennifer-richards (approved):

PR #1229, opened 2026-05-10
fix: glitch when Github call gets hash but download fails

rudimatz:
image

fixing the issue when the metadata result has a head_sha but the status is failed (when previous logic assumed that when the head_sha exists, its always a success)

### DIFF
--- a/client/app/pages/docs/[id]/publication-preparation.vue
+++ b/client/app/pages/docs/[id]/publication-preparation.vue
@@ -444,8 +444,8 @@ const fetchAndVerifyMetadata = async () => {
     step.value = {
       type: 'error',
       errorText: `Failed to validate metadata. Request status was still
-      ${JSON.stringify(resultsCreate.status)} after ${MAXIMUM_ATTEMPTS_DURATION_MS / 1000} seconds
-      and ${attemptCount} attempts.`,
+      ${JSON.stringify(resultsCreate.status)} after ${MAXIMUM_ATTEMPTS_DURATION_MS / 1000}
+      seconds.`,
       showResyncButton: true
     }
     return
--- a/client/app/pages/docs/[id]/publication-preparation.vue
+++ b/client/app/pages/docs/[id]/publication-preparation.vue
@@ -271,7 +271,7 @@ const { data: rfcToBe, error: rfcToBeError, status: rfcToBeStatus, refresh: rfcT
   }
 )
 
-const { data: metadataValidationResults, error: metadataValidationResultsError, status: metadataValidationResultsStatus } = await useAsyncData(
+const { data: metadataValidationResults, error: metadataValidationResultsError, status: metadataValidationResultsStatus, refresh: metadataValidationResultsRefresh } = await useAsyncData(
   () => `draft-${draftName.value}-metadata-validation-results`,
   () => api.documentsMetadataValidationResultsList({
     draftName: draftName.value,
@@ -456,10 +456,17 @@ const fetchAndVerifyMetadata = async () => {
 
 const deleteMetadataValidationAndRetry = async (headSha: string) => {
   step.value = { type: 'loading' }
+  // Refresh to get the current head_sha — the stored one may be stale if the
+  // validation task ran and set a head_sha after the page was first loaded.
+  await metadataValidationResultsRefresh()
+  const currentResult = metadataValidationResults.value as unknown
+  const currentHeadSha = isMetadataValidationResults(currentResult)
+    ? currentResult.headSha!
+    : headSha
   try {
     await api.metadataValidationResultsDelete({
       draftName: draftName.value,
-      headSha,
+      headSha: currentHeadSha,
     })
   } catch (error: unknown) {
     snackbarForErrors({ snackbar, error, defaultTitle: "Couldn't delete validation results" })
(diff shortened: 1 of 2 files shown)


rudimatz:
@jennifer-richards I discovered an edge case on my local:
my local celery was not working properly. when celery is not working, the getMetadata task will stay `pending` forever.
Our current logic doesn't allow to delete a pending MetadataValidationResult.

So this would need manual intervention, otherwise the user is stuck.

Simple solution could be to remove the guard and allow deleting pending records.
But this could possibly lead to deleting a request that runs all fine but is still in the ~5 second window of pending.

Since in the UI logic we always call the delete followed by a "re-fetch" I think this would be acceptable behavior.
Any preference @jennifer-richards

client/app/pages/docs/[id]/publication-preparation.vue
```diff
@@ -456,10 +458,17 @@ const fetchAndVerifyMetadata = async () => {
 
 const deleteMetadataValidationAndRetry = async (headSha: string) => {
   step.value = { type: 'loading' }
+  // Refresh to get the current head_sha — the stored one may be stale if the
+  // validation task ran and set a head_sha after the page was first loaded.
+  await metadataValidationResultsRefresh()
+  const currentResult = metadataValidationResults.value as unknown
```
Comment (holloway) [inline client/app/pages/docs/[id]/publication-preparation.vue]: Can `metadataValidationResults`'s type be changed to include unknown / undefined etc, rather than casting?

client/app/pages/docs/[id]/publication-preparation.vue
```diff
@@ -456,10 +458,17 @@ const fetchAndVerifyMetadata = async () => {
...
+  const currentHeadSha = isMetadataValidationResults(currentResult)
+    ? currentResult.headSha!
```
Comment (holloway) [inline client/app/pages/docs/[id]/publication-preparation.vue]:
```suggestion
  const currentHeadSha = isMetadataValidationResults(currentResult) && currentResult.headSha
    ? currentResult.headSha
```

Comment (holloway) [reply client/app/pages/docs/[id]/publication-preparation.vue]: This logic might not be quite right but the idea is to test the headSha rather than asserting`!` that it has a value

jennifer-richards:
> my local celery was not working properly. when celery is not working, the getMetadata task will stay `pending` forever. Our current logic doesn't allow to delete a pending MetadataValidationResult.
> 
> So this would need manual intervention, otherwise the user is stuck.

If celery is not working correctly in production, intervention will be needed anyway so I don't think I'm terribly worried about this.

I think that deleting the `MetadataValidationResult` when will allow multiple tasks to be queued. ~~When celery comes back online,~~  That _might_ be ok, but we'd need to think through the impact. It may lead to strange inconsistencies.

client/app/pages/docs/[id]/publication-preparation.vue
```diff
@@ -442,7 +442,9 @@ const fetchAndVerifyMetadata = async () => {
   if (resultsCreate.status !== 'success' && resultsCreate.status !== 'failed') {
     step.value = {
       type: 'error',
-      errorText: `Failed to validate metadata. Request status was still ${JSON.stringify(resultsCreate.status)}.`,
+      errorText: `Failed to validate metadata. Request status was still
+      ${JSON.stringify(resultsCreate.status)} after ${MAXIMUM_ATTEMPTS_DURATION_MS / 1000} seconds
+      and ${attemptCount} attempts.`,
```
Comment (jennifer-richards) [inline client/app/pages/docs/[id]/publication-preparation.vue]: The `attemptCount` is the count of attempts to poll status? If so, I'd probably leave that out of the error. It may cause confusion as to "attempts to do what?" and the elapsed time is really the important detail.

client/app/pages/docs/[id]/publication-preparation.vue
```diff
@@ -46,7 +46,7 @@
         <div :class="step.showDeleteAndRetryButton ? 'flex justify-between' : 'text-center'">
           <BaseButton v-if="step.showDeleteAndRetryButton" btn-type="outline"
             @click="deleteMetadataValidationAndRetry(step.showDeleteAndRetryButton.headSha)">
-            Retry metadata validation
+            Delete and re-fetch from repo
```
Comment (jennifer-richards) [inline client/app/pages/docs/[id]/publication-preparation.vue]: Is the meaning of "Delete" clear to the user here? My feeling is it's an internal detail about how we manage the validation data and it may cause the user to worry about what exactly they're deleting.

If you don't like the original label, maybe "Discard validation result and retry" or something?

jennifer-richards (changes requested):
Functionally seems ok, but I have a couple worries about the UI text changes. I think they may be asking the user to understand things about the internal workings that are more confusing than helpful to them. Happy to discuss if you disagree (particularly the bit about "Delete" because I'm not sure I understand what you were aiming to convey).

client/app/pages/docs/[id]/publication-preparation.vue
Comment (rudimatz) [reply client/app/pages/docs/[id]/publication-preparation.vue]: yes... it will be basically be 10 because we try once per second. I thought its nice to surface info we already have... but I can remove it if you consider it like "too much information"

client/app/pages/docs/[id]/publication-preparation.vue
Comment (rudimatz) [reply client/app/pages/docs/[id]/publication-preparation.vue]: yes, we basically showed 2 different "re-try" buttons on the UI and the difference is not clear. 
Using this button, we not only discard the validation, but we re-fetch the info from the repo and then compare the DB data with fresh data, so I thought its good to explain that.

jennifer-richards (approved):

PR #1231, opened 2026-05-11
fix: location of series info in xml

### LINKED ISSUE #1228: metadata validation looks for <seriesInfo> in the wrong context
When validating an RfcToBe destined to be a BCP, the metadata validation failed. It correctly showed that the database value for subseries was "BCP NNN" but showed the document's value as empty and flagged the mismatch.

It appears to me that the validation code is looking for `<seriesInfo>` elements in the root of the XML document. Per [RFC 7991](https://datatracker.ietf.org/doc/html/rfc7991#autoid-153) (and in the specific xml file being tested), `<seriesInfo>` is found inside the `<front>` element.

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1228

### DIFF
--- a/rpc/lifecycle/metadata.py
+++ b/rpc/lifecycle/metadata.py
@@ -585,25 +585,27 @@ class MetadataComparator:
 
     def compare_subseries(self):
         """Compare subseries field"""
-        xml_value = self.xml_metadata.get("subseries", [])
-        if not xml_value:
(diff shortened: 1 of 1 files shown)


rudimatz:
image

show string like `BCP 123` instead of JSON and simplify logic

jennifer-richards (approved):

PR #1235, opened 2026-05-11
feat: change default email

rudimatz:
use rfc-editor@rfc-editor.org as default for emails sent from purple


rudimatz:
fyi @ajeanmahoney @sginoza

rudimatz:
OBE

PR #1238, opened 2026-05-12
feat: public queue API shall return ALL action holders

rudimatz:
relates to https://github.com/ietf-tools/queue/issues/54#issuecomment-4426199910

wait for final confirmation


rudimatz:
confirmation given, we wanna show ALL actionholders on queue site (incl. the not-completed ones).

(UI on queue site can decide how the display them) @holloway

holloway (approved):

jennifer-richards (approved):

PR #1245, opened 2026-05-13
feat: add call to send queue to DT

rudimatz:
adding a flag to be able to disable it, and also a task to be able to schedule/call it via ADMIN


rudimatz:
set `PURPLE_NOTIFY_DT_QUEUE_ENABLED` to anything other than "false" to enable

jennifer-richards (approved):

PR #1249, opened 2026-05-14
feat: notify queue for drafts created, but not yet in_progress

### LINKED ISSUE #75: Enqueued doc does not show on demo queue site
Successfully enqueued draft-ietf-teas-rfc8776-update to demo purple, and closed the enqueuing assignment, but the I-D does not show on the demo queue site.

rudimatz:
simplify logic, apply DRY and remove N+1 queries.

use existing `in_queue()` filter to determine whether a draft should be considered for queue notification

fixes https://github.com/ietf-tools/queue/issues/75#issuecomment-4451611383

### DIFF
--- a/rpc/lifecycle/notifications.py
+++ b/rpc/lifecycle/notifications.py
@@ -43,7 +43,7 @@ def notify_queue_precompute():
 
 
 def get_updated_rfcs_since(current_check_time):
-    """Get IDs of in-queue RFCs updated since last check."""
+    """Return a queryset of in-queue RFCs updated since last check."""
 
     candidate_ids = set()
 
@@ -92,12 +92,7 @@ def get_updated_rfcs_since(current_check_time):
         .values_list("rfc_to_be", flat=True)
     )
 
-    # Single query to filter candidates to only in-queue documents
-    return set(
-        RfcToBe.objects.in_queue()
-        .filter(pk__in=candidate_ids)
-        .values_list("id", flat=True)
-    )
+    return RfcToBe.objects.in_queue().filter(pk__in=candidate_ids)
 
 
 def process_rfctobe_changes_for_queue():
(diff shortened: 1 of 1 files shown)


rpc/lifecycle/notifications.py
```diff
@@ -43,95 +43,66 @@ def notify_queue_precompute():
...
+        .values_list("rfc_to_be", flat=True)
+    )
+    candidate_ids.update(
+        FinalApproval.history.filter(history_date__gt=current_check_time)
+        .exclude(rfc_to_be=None)
+        .values_list("rfc_to_be", flat=True)
+    )
+
+    # Single query to filter candidates to only in-queue documents
+    return set(
```
jennifer-richards: Is the `set()` just here to avoid duplicates? You might consider using `.distinct()` instead and passing back a queryset. That's often a more reusable pattern than working with lists/sets of IDs

rpc/lifecycle/notifications.py
```diff
@@ -152,9 +123,7 @@ def process_rfctobe_changes_for_queue():
         recent_change_threshold = current_check_time - timezone.timedelta(minutes=1)
 
         # Check for recent changes - if changes happened in last minute, abort
-        recent_changes_exist = get_recent_changes(recent_change_threshold)
-
-        if recent_changes_exist:
+        if get_updated_rfcs_since(recent_change_threshold):
```
jennifer-richards: would be come a `.exists()` check if you were to adopt the suggestion to return a queryset

jennifer-richards (approved):
Approving, but one suggestion about adjusting the API to consider. (N.b. that I have not looked at what else you're doing with the method that might need to adjust if you changed the return value)

jennifer-richards (approved):

PR #1262, opened 2026-05-19
feat: check action holders to determine in-final-review

### LINKED ISSUE #54: Document Final Review page should also show Action Holders
When Action Holders are assigned, the final review page for the document should list them. For example, draft-ietf-scim-device-model has both AD and IANA action holders.

See https://queue.staging.rfc-editor.org/final-review/draft-ietf-scim-device-model/ vs https://www.rfc-editor.org/auth48/rfc9944

rudimatz:
fix https://github.com/ietf-tools/queue/issues/54#issuecomment-4480168446 and same behavior in purple "Final Review" page

keep existing behavior and add new filter `pending_final_review` that takes both ActionHolders and FinalApprovals into account.

I didn't update the behavior of the existing `pending_final_approval` filter because that would have been semantically incorrect

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -681,9 +681,10 @@ class QueueFilter(django_filters.FilterSet):
     pending_final_review = django_filters.BooleanFilter(
         method="filter_pending_final_review",
         help_text="Filter by pending final review status. True returns drafts with "
-        "at least one pending author approval (FinalApproval) or uncompleted action "
-        "holder. False returns drafts where all author approvals are done and all "
-        "action holders are completed.",
+        "at least one pending author approval (FinalApproval) or at least one "
+        "uncompleted action holder. False returns drafts where at least one author "
+        "approval exists, all author approvals are done, and no action holders are "
+        "uncompleted.",
     )
 
     def filter_pending_final_approval(self, queryset, name, value):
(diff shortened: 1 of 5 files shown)


rudimatz:
once merged, **queue** site should use the new API client filters `pendingFinalReview` just like purple @holloway

rpc/api.py
```diff
@@ -678,6 +678,13 @@ class QueueFilter(django_filters.FilterSet):
         "at least one pending final approval, false returns drafts where all final "
         "approvals are approved.",
     )
+    pending_final_review = django_filters.BooleanFilter(
+        method="filter_pending_final_review",
+        help_text="Filter by pending final review status. True returns drafts with "
+        "at least one pending author approval (FinalApproval) or uncompleted action "
+        "holder. False returns drafts where all author approvals are done and all "
+        "action holders are completed.",
```
jennifer-richards: I think the `False` case is `where at least one author approval exists, all author approvals are done, and all action holders (if any) are competed` - is that accurate?

jennifer-richards (approved):
Approved, but you may want to adjust one comment if I didn't misread the code.

rpc/api.py
rudimatz: yes, comment updated

holloway:
> once merged, queue site should use the new API client filters pendingFinalReview just like purple

thanks @rudimatz this is now done

PR #1268, opened 2026-05-20
feat: mark enqueuer assignment as done once in queue

rudimatz:
move logic from FE to BE.

create endpoint to move draft disposition from `created` to `in_progress` and add the person doing it as `enqueuer` with state `done`
(same behavior as done in xfer: once in queue, we consider enqueuer job done)

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -1331,11 +1331,11 @@ class RfcToBeViewSet(viewsets.ModelViewSet):
         enqueuer_role = RpcRole.objects.get(slug="enqueuer")
         rpc_person = request.user.rpcperson()
         if rpc_person is not None:
-            Assignment.objects.create(
+            Assignment.objects.update_or_create(
                 rfc_to_be=rfctobe,
                 role=enqueuer_role,
                 person=rpc_person,
-                state=Assignment.State.DONE,
+                defaults={"state": Assignment.State.DONE},
             )
         return Response(RfcToBeSerializer(rfctobe, context={"request": request}).data)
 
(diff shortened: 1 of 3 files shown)


holloway (approved):

rpc/api.py
```diff
@@ -1311,6 +1311,34 @@ def publish(self, request, draft__name=None):
...
+            raise serializers.ValidationError(
+                f"Cannot enqueue: disposition is '{rfctobe.disposition_id}', "
+                "expected 'created'."
+            )
+        rfctobe.disposition_id = DispositionName.IN_PROGRESS
+        rfctobe.save()
+        enqueuer_role = RpcRole.objects.get(slug="enqueuer")
+        rpc_person = request.user.rpcperson()
+        if rpc_person is not None:
+            Assignment.objects.create(
```
jennifer-richards: Is there any other path that could result in an enqueuer `Assignment`? Even if it's not expected, using an `update_or_create()` here would guard against that.

Also, this is going to create an `Assignment` in the `DONE` state. Do we (or will we) have stats or other processing that might expect for there to be a history entry transitioning from in-progress to done? If so, for consistency it might be preferable to create one as in progress before taking the action, then immediately transition to done.

(If you were to go that route, I'd think you'd want to create the in-progress assignment at the start (in case it fails for some reason), then adjust the `rfctobe` state, and then change the assignment to done.)

jennifer-richards (changes requested):
See inline. I think using `update_or_create()` is probably important. The rest of my comment is just a suggestion to consider. It might be more complex than necessary.

rpc/api.py
rudimatz: there is no other path really, and we have a guard that the draft is NOT yet enqueued. but I agree using `update_or_create` is more defensive

jennifer-richards (approved):

PR #1271, opened 2026-05-20
feat: add filter for iana status

rudimatz:
adding new filter for IANA status
image

### DIFF
--- a/client/app/pages/queue/queue.vue
+++ b/client/app/pages/queue/queue.vue
@@ -68,7 +68,7 @@
             <select v-model="selectedIanaStatusFilter" :class="SELECT_STYLE">
               <option :value="null">All Statuses</option>
               <option v-for="status in allIanaStatuses" :key="status.slug" :value="status.slug">
-                {{ status.name }}
+                {{ status.label }}
               </option>
             </select>
           </div>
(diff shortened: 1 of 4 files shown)


rudimatz:
relates to https://github.com/ietf-tools/rfced-xfer/issues/230

client/app/pages/queue/queue.vue
```diff
@@ -59,6 +59,20 @@
...
           </div>
         </fieldset>
+        <fieldset>
+          <legend class="font-bold text-sm flex items-end">
+            IANA Status
+            <span class="text-md">&nbsp;</span>
+          </legend>
+          <div class="flex flex-col pt-1">
+            <select v-model="selectedIanaStatusFilter" :class="SELECT_STYLE">
+              <option :value="null">All Statuses</option>
```
holloway, line 69: I'm not sure that an `<option>` value can literally be `null` (I thought they were `undefined` or a `string`) but if it works I guess it's ok

holloway (approved):

client/app/pages/queue/queue.vue
jennifer-richards: Worries me a little that it might seem to work but have a hidden bug

jennifer-richards (approved):
Echo Matthew's concern about the null option, but if you're certain it's working (including if you select a value in the new input, then reset it to the default), it is fine by me

client/app/pages/queue/queue.vue
rudimatz: it **does** work. What solution would you find cleaner, empty string?

rudimatz:
will merge after further checks with Jennifer.

Generated code: (no `"null"` string )
image

client/app/pages/queue/queue.vue
jennifer-richards: Rudi showed me - it looks like it turns this into an `<option>All Statuses</option>` that has JS magic attached. Seems to work, so good enough for me.

PR #1272, opened 2026-05-20
feat: filter by assignment in final review

### LINKED ISSUE #702: Should not be able to put a doc without an RFC number and a final review editor into final review
Currently, just adding approvers causes a document to move into Final Review. This can happen even though the document does not have an RFC number nor a final review editor.

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/702

current logic requires Action Holder to be **present**, and completed.
Expected logic: either no ActionHolder or all completed

Additional filter: we ALWAYS expect that draft has at least a final review editor assignment

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -703,9 +703,10 @@ class QueueFilter(django_filters.FilterSet):
         return queryset
 
     def filter_pending_final_review(self, queryset, name, value):
+        has_fre = queryset.filter(assignment__role__slug="final_review_editor")
         if value is True:
             # has at least one pending FinalApproval OR an uncompleted ActionHolder
-            return queryset.filter(
+            return has_fre.filter(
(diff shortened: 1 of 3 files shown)


rpc/api.py
```diff
@@ -703,21 +704,27 @@ def filter_pending_final_approval(self, queryset, name, value):
         return queryset
 
     def filter_pending_final_review(self, queryset, name, value):
+        has_fre = queryset.filter(assignment__role__slug="final_review_editor")
```
jennifer-richards: Should this exclude inactive assignments? 

In particular, if a final review editor is accidentally assigned early and the assignment is terminated/canceled, I don't think we want this to select it.

jennifer-richards (changes requested):
One question/suggestion inline

jennifer-richards (approved):

PR #1278, opened 2026-05-21
feat: show labels in final review sections

### LINKED ISSUE #1275: Add markdown/github labels to Final Review view
It would be helpful to display the markdown and github labels to the [internal Final Review page](https://purple.rfc-editor.org/final-review) so editors can easily know which process is being used.  It might be helpful to have an explicit xml label as well (which would apply to anything that is not marked as markdown).

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1275

### DIFF
--- a/client/app/components/FinalReviewsDone.vue
+++ b/client/app/components/FinalReviewsDone.vue
@@ -8,6 +8,7 @@
       <colgroup>
         <col class="w-8">
         <col>
+        <col class="w-40">
         <col class="w-28">
         <col class="w-28">
         <col class="w-96">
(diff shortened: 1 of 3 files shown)


holloway (approved):

jennifer-richards (approved):
I don't think this adds the xml label that was suggested, might see if Sandy wants that to be added as a feature request to consider

rudimatz:
@jennifer-richards added comment https://github.com/ietf-tools/purple/issues/1275#issuecomment-4518725286

PR #1279, opened 2026-05-21
feat: no error message for published docs without draft

rudimatz:
as discussed, no need to show warning if already published


jennifer-richards (approved):
Should other terminal states be excluded too?

rudimatz:
@jennifer-richards valid point. but there are no docs without draft and not published. So I think in this case its valid to keep the exclude filter tight.

PR #1284, opened 2026-05-22
feat: improve the labels management

rudimatz:
image

separate in use and not in use

(added edit options for the `not in used`, to be able to switch to "in use")
as requested by @sginoza

### DIFF
--- a/client/app/components/RpcLabelEditDialog.vue
+++ b/client/app/components/RpcLabelEditDialog.vue
@@ -98,13 +98,13 @@
 
         <div class="sm:grid sm:grid-cols-3 sm:items-start sm:gap-4 sm:py-6">
           <label for="is-used" class="block text-sm font-medium leading-6 text-gray-900 sm:pt-1.5">
-            May be Used
+            Is Assignable
           </label>
           <div class="mt-2 sm:col-span-2 sm:mt-0">
             <input
               id="is-used" v-model="label.used" name="is-used" type="checkbox"
               class="h-4 w-4 rounded border-gray-300 text-indigo-600 focus:ring-indigo-600">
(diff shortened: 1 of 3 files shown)


client/app/components/RpcLabelEditDialog.vue
```diff
@@ -37,7 +37,7 @@
               <input
                 id="slug" v-model="label.slug" type="text" name="slug"
                 class="block flex-1 border-0 bg-transparent py-1.5 pl-1 text-gray-900 placeholder:text-gray-400 focus:ring-0 sm:text-sm sm:leading-6"
-                placeholder="happy tree">
+                placeholder="e.g. `markdown`">
```
jennifer-richards: use of backticks in the placeholder text intentional? Slightly odd usage.

jennifer-richards (commented):
The code seems fine, but I'm a bit wary about introducing the "delete" option and about the somewhat confusing label "In use".

I think the majority of the not-"In use" labels are actually applied to documents, and in the informal sense, are "In use." They cannot be deleted, but the UI will make it appear that they can. We'd really want the back-end to tell the front-end whether the label is currently deletable, rather than making the user guess.

Is there an immediate need for the delete option, or is it a "nice-to-have"? I think we should discuss the deletion a bit more widely before adding it as a feature.

client/app/components/RpcLabelEditDialog.vue
rudimatz: it was intentional, but changed to single quote

rudimatz:
after offline discussion, I removed the `delete` option which requires further refinement of requirements.

jennifer-richards (approved):

PR #1295, opened 2026-05-25
fix: wrong labels for action holders

rudimatz:
field descriptions are confusing and lead to some wrong input by editors


jennifer-richards (changes requested):
fwiw, I find "Body" to be ambiguous and unclear. It makes more sense than what was there before, though. Maybe it is better understood by the RPC than by me.

Marking request changes because I think it should explicitly show that the field is optional (if that's still true)

rudimatz:
after discussion with @ajeanmahoney I changed it to this
image

jennifer-richards (approved):

PR #1302, opened 2026-05-27
fix: publication email template

### LINKED ISSUE #1301: Fix email template "Publication"
The following text is hardcoded and needs to be made dynamic

```
INFORMATIONAL: This memo provides information for the Internet community.
It does not specify an Internet standard of any kind. Distribution of
this memo is unlimited.
```

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1301


jennifer-richards (approved):
looks ok, but maybe let sandy/alice weigh in before running with it?

Issue #1303, opened 2026-05-27
allow creation of ActionHolder for "body", no person selected


rudimatz:
as discussed, we will set the API to allow receiving ActionHolder creation without user, and the backend shall set the System user

(ActionHolder model is set to require a `dt_person` to be set)

relates to https://github.com/ietf-tools/purple/pull/1295

rudimatz:
fixed via https://github.com/ietf-tools/purple/issues/1303

PR #1311, opened 2026-05-28
feat: add toggle action holder and system user default

rudimatz:
add toggle for action holder to set EITHER person OR body

image

image

display in list either... or

added new field in serializer "display_name" which automatically returns body or person name


jennifer-richards (approved):
lgtm

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/1303

PR #1329, opened 2026-06-03
fix: API client generator error for queue count

rudimatz:
fix
```
Error: using @extend_schema on viewset class QueueCounts with parameters operation_id or operation will most likely result in a broken schema.
Errors:   1 (1 unique)
If not set, add @ts-nocheck in runtime.ts to avoid type errors from generated code
```


rudimatz:
code works as it is, but this will remove error from generator

jennifer-richards (approved):

PR #1333, opened 2026-06-03
feat: add task to backfill final_review start dates

### LINKED ISSUE #1318: Final Review page shows purple cutover date rather than date that final review started
For docs that were in AUTH48 during cutover. For example, draft-ietf-stir-servprovider-oob shows May 20, 2026, but the document entered AUTH48 on October 21, 2025.

rudimatz:
resolves https://github.com/ietf-tools/purple/issues/1318

start dates have been compiled from legacy DB (`AUTH48` state change) and validated.
Task re-writes history and updates the create-date of the `final-review-editor` assignment

### DIFF
--- a/rpc/lifecycle/final_review.py
+++ b/rpc/lifecycle/final_review.py
@@ -1,0 +1,134 @@
+# Copyright The IETF Trust 2026, All Rights Reserved
+
+import datetime
+import logging
+
+from django.utils import timezone
+
+logger = logging.getLogger(__name__)
+
+# (draft_name, start_date) — start_date is when the final_review_editor assignment was
+# created
+FINAL_REVIEW_START_DATES = [
+    ("draft-ietf-tls-rfc8446bis", "2025-12-15"),
+    ("draft-ietf-tls-keylogfile", "2025-12-16"),
+    ("draft-ietf-tls-tls12-frozen", "2026-01-05"),
+    ("draft-ietf-uta-require-tls13", "2026-01-06"),
+    ("draft-ietf-stir-servprovider-oob", "2025-10-21"),
+    ("draft-ietf-pce-pceps-tls13", "2026-01-16"),
+    ("draft-ietf-netconf-over-tls13", "2026-01-16"),
+    ("draft-ietf-lamps-rfc5019bis", "2026-01-16"),
+    ("draft-ietf-pce-sid-algo", "2026-02-19"),
+    ("draft-ietf-cose-merkle-tree-proofs", "2026-03-06"),
+    ("draft-ietf-scitt-architecture", "2026-03-06"),
+    ("draft-ietf-tls-hybrid-design", "2026-04-03"),
+    ("draft-ietf-pquip-hybrid-signature-spectrums", "2026-04-03"),
+    ("draft-ietf-pquip-pqc-engineers", "2026-04-27"),
+    ("draft-ietf-emu-bootstrapped-tls", "2026-04-27"),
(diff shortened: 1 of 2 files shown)


rpc/lifecycle/final_review.py
```diff
@@ -0,0 +1,142 @@
...
+    from rpc.models import Assignment, RfcToBe
+
+    HistoricalAssignment = Assignment.history.model
+    added = 0
+    skipped = 0
+
+    for draft_name, date_str in FINAL_REVIEW_START_DATES:
+        start_date = timezone.make_aware(
+            datetime.datetime.strptime(date_str, "%Y-%m-%d").replace(hour=12)
+        )
```
jennifer-richards: I'd suggest `.replace(hour=12, tzinfo=datetime.UTC)` instead of the `make_aware`. We want to be explicit about our timezone choice - we specifically want UTC for this purpose (even though it's hard to imagine we'd ever change Django's timezone).

e.g., on datatracker, which for Reasons still has its django time zone set to PST8PDT
```
>>> str(timezone.make_aware(datetime.datetime(2026,6,5).replace(hour=12)))
'2026-06-05 12:00:00-07:00'
```
vs
```
>>> str(datetime.datetime(2026,6,5).replace(hour=12, tzinfo=datetime.UTC))
'2026-06-05 12:00:00+00:00'
```

rpc/lifecycle/final_review.py
```diff
@@ -0,0 +1,142 @@
...
+        if not assignments.exists():
+            logger.warning(
+                "backfill_final_review_history: no final_review_editor assignment "
+                "for %s",
+                rfctobe.name,
+            )
+            skipped += 1
+            continue
+
+        for assignment in assignments:
```
jennifer-richards: I think there should only be one final review assignment, shouldn't there? If there's more than one, I think the code would need to indicate which one the start date applies to, so should cause a skip.

rpc/lifecycle/final_review.py
```diff
@@ -0,0 +1,142 @@
...
+            existing = HistoricalAssignment.objects.filter(
+                id=assignment.id, history_type="+"
+            ).first()
```
jennifer-richards: Still in paranoia mode, I'd probably guard against > 1 match here. I think it's impossible under normal operation, but easy enough to check.

rpc/lifecycle/final_review.py
```diff
@@ -0,0 +1,142 @@
...
+                    "backfill_final_review_history: %supdated history_date for %s "
+                    "assignment #%s → %s",
+                    prefix,
+                    rfctobe.name,
+                    assignment.id,
+                    date_str,
+                )
+            else:
+                if not dry_run:
+                    HistoricalAssignment.objects.create(
```
jennifer-richards: Do we expect this code path to be exercised? It worries me that we'd need it, and I think it'd be safer to abort if we hit it because I think it means we don't understand our database contents.

jennifer-richards (changes requested):
Top level comment: If we can simplify this and have it fail if the database isn't in the state we expect, that'd be preferable. E.g., see the comment on the "HistoricalAssignment doesn't exist" branch.

A few comments / change requests inline. Additionally, a comment should be added to the RfcToBe histories recording that the history has been adjusted.

rudimatz:
added new guard, mapping table contains AssignmentHistory PKs, validate against them to make sure the right entry is updated.
also add history comment to indicate the value was updated with backfill
image

jennifer-richards (approved):

PR #1335, opened 2026-06-04
feat: make rev of RfcToBe editable

### LINKED ISSUE #1327: Document Info tab: Revision should be editable
The Revision field captures the revision number of the I-D when it was enqueued. However, new revisions are provided by authors after the document has been enqueued. It would be helpful if this field was editable so we can note the new version.

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1327

add new field `RfcToBe.rev`. If set, it will overwrite the draft's original value `draft.rev`
when writing, only update `RfcToBe.rev`, never touch `draft.rev`

### DIFF
--- a/client/app/components/DocInfoCard.vue
+++ b/client/app/components/DocInfoCard.vue
@@ -64,7 +64,11 @@
         </DescriptionListItem>
         <DescriptionListItem term="Revision" :spacing="spacing">
           <DescriptionListDetails>
-            <span class="font-mono">{{ rfcToBe.draft?.rev ?? '(none)' }}</span>
+            <PatchRfcToBeField fieldName="rev" :is-read-only="props.isReadOnly"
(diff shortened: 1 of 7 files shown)


jennifer-richards (commented):
Couple questions:

1. Would it be simpler to set `RfcToBe.rev` on every import and not deal with the fallback to draft when it's blank? Is there some reason not to like this?
2. Should the UI more prominently show that the `RfcToBe` was created at a different rev than is currently selected?
3. Does the change show up in the doc history? (It should, and I expect it does)

rudimatz:
as discussed with @jennifer-richards , might be better to not implement `RfcToBe.rev` as fallback, but as proper value. This would need a migration script to copy over values from `draft.rev`.

needs further though, put on hold for now

rudimatz:
final solution is not to use it as fallback, but independent field (same logic as `pages`)
It will be filled on import, and will be updatable via doc info page

Migration script also contains function to initially fill the `rev`

jennifer-richards (approved):

PR #1356, opened 2026-06-18
feat: withdraw doc 

### LINKED ISSUE #1348: Can't withdraw document from the queue
There used to be the ability to withdraw a document from the queue by selecting "Withdrawn" from the Disposition field on the document's Info tab (see https://github.com/ietf-tools/purple/issues/720). This feature is not available now.

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/1348
adds dedicated "Withdraw@ button in UI, displayed only for active drafts

in withdraw dialog, the user can also add a optional `comment` for the withdrawal

### DIFF
--- a/client/app/components/DocHeader.vue
+++ b/client/app/components/DocHeader.vue
@@ -34,6 +34,13 @@
               <Icon name="ei:spinner-3" size="1rem" class="animate-spin" />
             </span>
           </BaseButton>
+          <BaseButton
+            v-if="props.rfcToBe?.disposition === 'in_progress'"
+            @click="isWithdrawDialogShown = true"
+            class="flex items-center !bg-rose-600 hover:!bg-rose-500"
+          >
+            <span>Withdraw</span>
+          </BaseButton>
           <BaseButton @click="openAssignmentFinishedModal" class="flex items-center">
             <span>Manage Assignments</span>
             <span v-if="isLoadingFinishAssignmentsModal" class="w-3">
@@ -44,6 +51,12 @@
       </div>
     </div>
   </header>
+
+  <WithdrawDraftDialog
+    v-model:is-shown="isWithdrawDialogShown"
+    :draft-name="props.draftName"
+    @success="emit('withdrawn')"
+  />
 </template>
 
 <script setup lang="ts">
@@ -58,7 +71,9 @@ type Props = {
 }
 
 const props = defineProps<Props>()
-const emit = defineEmits<{ assignmentsChanged: [] }>()
+const emit = defineEmits<{ assignmentsChanged: []; withdrawn: [] }>()
+
+const isWithdrawDialogShown = ref(false)
 
 const overlayModal = inject(overlayModalKey)
 const isAprilFirst = computed(() => props.rfcToBe?.isAprilFirstRfc === true)
--- a/client/app/components/DocHeader.vue
+++ b/client/app/components/DocHeader.vue
@@ -36,8 +36,9 @@
           </BaseButton>
           <BaseButton
             v-if="props.rfcToBe?.disposition === 'in_progress'"
+            btn-type="delete"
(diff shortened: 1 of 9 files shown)


rudimatz:
image

client/app/components/DocHeader.vue
```diff
@@ -34,6 +34,13 @@
               <Icon name="ei:spinner-3" size="1rem" class="animate-spin" />
             </span>
           </BaseButton>
+          <BaseButton
+            v-if="props.rfcToBe?.disposition === 'in_progress'"
+            @click="isWithdrawDialogShown = true"
+            class="flex items-center !bg-rose-600 hover:!bg-rose-500"
```
holloway: can you use btn-type prop (mapped to `classForBtnType`) rather than these one-off styles?

client/app/components/DocHeader.vue
```diff
@@ -58,7 +71,9 @@ type Props = {
 }
 
 const props = defineProps<Props>()
-const emit = defineEmits<{ assignmentsChanged: [] }>()
+const emit = defineEmits<{ assignmentsChanged: []; withdrawn: [] }>()
```
holloway, line 75: Is the type really an empty array or will it have a length during runtime?

afaik emits have broadcasters and receivers so it's probably worth having a shared type on both ends, from a utils file

client/app/components/WithdrawDraftDialog.vue
```diff
@@ -0,0 +1,113 @@
...
+                        rows="3"
+                        class="block w-full rounded-md border-0 py-1.5 text-gray-900 dark:text-neutral-100 dark:bg-neutral-800 shadow-sm ring-1 ring-inset ring-gray-300 dark:ring-neutral-600 placeholder:text-gray-400 focus:ring-2 focus:ring-inset focus:ring-indigo-600 sm:text-sm sm:leading-6"
+                        placeholder="Add a comment (optional)..."
+                      />
+                    </div>
+                  </div>
+                </div>
+              </div>
+              <div class="bg-gray-50 dark:bg-neutral-800 px-4 py-3 sm:flex sm:flex-row-reverse sm:px-6 gap-2">
+                <button
```
holloway: Any reason not to use BaseButton for these?

client/app/components/WithdrawDraftDialog.vue
```diff
@@ -0,0 +1,113 @@
...
+    })
+  } catch (error: unknown) {
+    snackbarForErrors({ snackbar, defaultTitle: 'Withdrawing draft failed', error })
+    isSubmitting.value = false
+    return
+  }
+
+  // Withdraw committed — the comment is a best-effort annotation.
+  try {
+    if (comment.value.trim()) {
```
holloway: use a tmp var for this trimmed value so it doesn't get out of sync with the duplicate trimmed value?

client/app/components/WithdrawDraftDialog.vue
```diff
@@ -0,0 +1,113 @@
...
+                <button
+                  type="button"
+                  class="inline-flex w-full justify-center rounded-md bg-rose-600 px-3 py-2 text-sm font-semibold text-white shadow-sm hover:bg-rose-500 disabled:opacity-50 sm:w-auto"
+                  :disabled="isSubmitting"
+                  @click="handleWithdraw"
+                >
+                  <Icon v-if="isSubmitting" name="ei:spinner-3" size="1.1em" class="animate-spin mr-1" />
+                  Withdraw
+                </button>
+                <button type="button" class="btn-secondary" :disabled="isSubmitting" @click="cancel">Cancel</button>
```
holloway: If the screenshot is still accurate then perhaps we might need more spacing between the cancel and withdraw buttons? Obv don't want people clicking something accidentally.

holloway (approved):
LGTM just minor feedback

PR #1357, opened 2026-06-19
fix: reference error in cluster UI

rudimatz:
fixing error

```

index.vue:50 Uncaught (in promise) ReferenceError: ref is not defined
    at setup
    
    ```


rudimatz:
@holloway this is to fix a local error I got, seems to be related to Nuxt's auto-import.
This fix worked for me, but maybe you see a better way to resolve.

holloway (approved):
I'm ok with regular imports (seems harmless), but I'm also guessing there are 100 other places that use `ref` without importing it.

Typically when I've seen autoimport bugs restarting the dev server fixes it but I'm guessing you tried that and it didn't work.

Anyway, yeah this seems harmless.

PR #1359, opened 2026-06-22
feat: update subject line for final review email

### LINKED ISSUE #1343: update subject line of "Final Review" mail
When the user clicks 'Send Email' then 'Final Review', please update the subject line as follows. (Currently this is being manually updated for each document moved to final review.)
current: 
RFC-to-be NNNNN draft-foo for your review
new:  
Final Review: RFC-to-be NNNNN (draft-foo) in XML

For example:
current:
RFC-to-be 10007 draft-ietf-lamps-keyusage-crl-validation for your review
new: 
Final Review: RFC-to-be 10007 (draft-ietf-lamps-keyusage-crl-validation) in XML

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1343

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -2497,7 +2497,8 @@ class RfcMailTemplatesList(views.APIView):
                 "cc": list(interested_parties),
             },
             "finalreview": {
-                "subject": f"Final Review: RFC-to-be {rfc_number} {draft_name} in XML",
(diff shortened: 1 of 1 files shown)


jennifer-richards (approved):
Approving, but noting for the record that this will probably require manual editing for April 1 RFCs or any RFC that does not have a draft. In those cases, the subject will appear like `Final Review: RFC-to-be 12345 (RFC 12345) in XML`. The editor can manually adjust that for now.

PR #1364, opened 2026-06-23
feat: update blocking rules, add gate 0

### LINKED ISSUE #1330: "Tools Issue" label should be blocking
When the "Tools Issue" label is assigned, it should block the document.

### LINKED ISSUE #1332: "Author Input Required" label should always block
Currently the "Author Input Required" label only creates a block if the document is not in Final Review. When the document is in Final Review, applying the label does not create a block, but it should always create a block.

### LINKED ISSUE #1337: Cluster management: draft-ietf-teep-otrp-over-http should be blocked
It looks like draft-ietf-teep-otrp-over-http should be blocked, but it shows as available for assignment. 

I'm seeing these in the ref field: 
refqueue draft-ietf-teep-protocol (in queue)
Not received draft-ietf-suit-update-management

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1332 and fix https://github.com/ietf-tools/purple/issues/1337 and fix https://github.com/ietf-tools/purple/issues/1330

Update (and simplify) block logic following updated rulesets

### DIFF
--- a/rpc/lifecycle/blocked_assignments.py
+++ b/rpc/lifecycle/blocked_assignments.py
@@ -33,33 +33,37 @@ def get_block_reasons(rfc: RfcToBe) -> set[str]:
     """Compute whether blocked and collect blocking reasons."""
     reasons: set[str] = set()
 
+    # Gate 0: Always blocks regardless of current assignment
+    if rfc.labels.filter(slug="Author Input Required").exists():
+        reasons.add(BlockingReason.LABEL_AUTHOR_INPUT_REQUIRED)
+    if rfc.labels.filter(slug="Stream Hold").exists():
+        reasons.add(BlockingReason.LABEL_STREAM_HOLD)
+    if rfc.labels.filter(slug="Tools Issue").exists():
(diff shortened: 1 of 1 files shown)


rpc/lifecycle/blocked_assignments.py
```diff
@@ -33,33 +33,37 @@ def get_block_reasons(rfc: RfcToBe) -> set[str]:
     """Compute whether blocked and collect blocking reasons."""
     reasons: set[str] = set()
 
+    # Gate 0: Always blocks regardless of current assignment
+    if rfc.labels.filter(slug="Author Input Required").exists():
```
jennifer-richards, line 37: Aside: we may want to look at how we're using `slug`. It's intended to be a URL-friendly short name without spaces/punctuation (except for hyphens/underscores). I don't think it's urgent, but might be something to revisit.

https://docs.djangoproject.com/en/6.0/glossary/#term-slug

jennifer-richards (approved):
One comment inline, but it's not a blocker. (See what I did there? 😁)

PR #1371, opened 2026-06-30
fix: improve author name extraction for metadata sync

### LINKED ISSUE #1368: Cannot update metadata from XML if author elements have only fullName / asciiFullName attributes
While doing the publication steps for 10007.
From https://purple.rfc-editor.org/docs/draft-ietf-lamps-keyusage-crl-validation/publication-preparation
- clicked 'Fetch and verify metadata'
- it returns 2 differences (the non-ascii organization name and the abstract. screenshot below.)
- clicked 'Update database to match document'
- still spinning 5+ minutes later. 

perhaps the spinning is due to "SECOM CO., LTD."  vs. "セコム株式会社"

view-source:https://datatracker.ietf.org/doc/draft-ietf-lamps-keyusage-crl-validation/doc.json
      "affiliation": "SECOM CO., LTD."

vs. https://github.com/rfc-editor-drafts/rfc10007/blob/main/rfc10007.xml
```
<organization ascii="SECOM CO., LTD." showOnFrontPage="true">セコム株式会社</organization>
```

[Thinking aloud] Perhaps, if it's not possible for the "database" to  match the "document" for this non-ascii org, then it could fall back to the document's ascii attribute in order to match. 

Image

jennifer-richards:
Author name extraction code had diverged between the `MetadataComparator` comparison and the metadata sync. This extracts the comparator's code and uses it in both contexts.

Tests the extracted method but does not tackle testing the comparator or sync code. That should be done as a separate project.

Fixes #1368 (I believe)

### DIFF
--- a/rpc/lifecycle/metadata.py
+++ b/rpc/lifecycle/metadata.py
@@ -291,7 +291,10 @@ class Metadata:
     def extract_name_from_author_dict(author_dict) -> str:
         """Extract name from an author_dict
 
-        Extracts the name as it should appear in the titlepage block of the RFC.
+        Extracts the name, formatted as for titlepage_name in datatracker. This should
+        match the Latinized form of the name in initials + surname format for now.
+        It's likely this will eventually change to capture non-Latin names, but we're
+        not doing that yet.
         """
         xml_name = (
             author_dict.get("initials", "").rstrip()
--- a/rpc/lifecycle/metadata.py
+++ b/rpc/lifecycle/metadata.py
@@ -293,7 +293,9 @@ class Metadata:
         Extracts the name as it should appear in the titlepage block of the RFC.
         """
         xml_name = (
-            author_dict.get("initials", "") + " " + author_dict.get("surname", "")
+            author_dict.get("initials", "").rstrip()
+            + " "
+            + author_dict.get("surname", "").lstrip()
         ).strip()
         if not xml_name:
             xml_fullname = (
--- a/rpc/lifecycle/metadata.py
+++ b/rpc/lifecycle/metadata.py
@@ -6,7 +6,6 @@ import logging
 import re
 import xml.etree.ElementTree as ET
 from itertools import zip_longest
-from typing import Any
 
 from django.db import transaction
 
@@ -96,8 +95,8 @@ class Metadata:
             "subseries": subseries,
         }
 
-    @staticmethod
-    def update_metadata(rfctobe, metadata):
+    @classmethod
+    def update_metadata(cls, rfctobe, metadata):
         """Update metadata fields from metadata dictionary
 
         First compares all fields to see what needs to be updated,
@@ -200,11 +199,7 @@ class Metadata:
                                 )
 
                             # Verify names match
-                            xml_name = (
-                                xml_author.get("initials", "")
-                                + " "
-                                + xml_author.get("surname", "")
-                            ).strip()
+                            xml_name = cls.extract_name_from_author_dict(xml_author)
                             db_name = db_author.titlepage_name
                             if xml_name != db_name:
                                 raise ValueError(
--- a/rpc/lifecycle/metadata.py
+++ b/rpc/lifecycle/metadata.py
@@ -6,6 +6,7 @@ import logging
 import re
 import xml.etree.ElementTree as ET
 from itertools import zip_longest
+from typing import Any
 
 from django.db import transaction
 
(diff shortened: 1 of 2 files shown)


rpc/lifecycle/metadata.py
```diff
@@ -290,6 +287,31 @@ def update_metadata(rfctobe, metadata):
...
+        Extracts the name as it should appear in the titlepage block of the RFC.
+        """
+        xml_name = (
+            author_dict.get("initials", "").rstrip()
+            + " "
+            + author_dict.get("surname", "").lstrip()
+        ).strip()
+        if not xml_name:
+            xml_fullname = (
+                author_dict.get("asciiFullname", "") or author_dict.get("fullname", "")
```
kesara, line 306: I think `fullname` have the priority here. `asciiFullname` can be used when `fullname` has non Latin characters.

rpc/lifecycle/tests_metadata.py
```diff
@@ -0,0 +1,69 @@
...
+            ),
+            "A. B. Clyde",
+            "initials+surname have priority",
+        )
+        self.assertEqual(
+            Metadata.extract_name_from_author_dict(
+                {"fullname": "Diane Egawa", "asciiFullname": "Frank Gouda"}
+            ),
+            "F. Gouda",
+            "asciiFullname has priority",
```
kesara, line 46: `xml2rfc` seems to prioritize `fullname` over `asciiFullname`. What's the origin of this logic?

kesara (changes requested):
Couple of questions about `fullname` vs `asciiFullname` priority.

rpc/lifecycle/tests_metadata.py
jennifer-richards: It was the existing code in purple, that is unchanged aside from extracting it to a method and adjusting the whitespace stripping

kesara: If you want to replicate `xml2rfc` logic this should be other way around.

kesara: Another thing to consider is when `fullname` is non latin and `asciiFullname` is present.
This is an example of that: https://www.rfc-editor.org/rfc/rfc9005.txt
```<author fullname="李呈" asciiFullname="Cheng Li">```
shows tile page name as `李呈 (C. Li)`.

Not really sure these are valid for this particular use case.

Comment (jennifer-richards) [reply rpc/lifecycle/tests_metadata.py:46]: Looking at the [txt output](https://raw.githubusercontent.com/rfc-editor-drafts/rfc10007/e499308f775e294216940d3e105a99bb0c0b9fc6/rfc10007.txt) from the files involved in the original report, you're right that xml2rfc uses fullname though it also includes the asciiFullname afterwards.

There may be an issue with titlepage name falling out from this, looking to see whether it's going to behave correctly (or at least consistently with other RFCs) as far as showing ascii vs non-ascii unicode titlepage names.

Comment (jennifer-richards) [reply rpc/lifecycle/tests_metadata.py:46]: Datatracker does not appear to have any non-latin `titlepage_name` authors for now. It probably should be able to display the non-latin form like the RFC does, but that will be a new feature. (There are some non-ascii names, such as the first author of https://datatracker.ietf.org/doc/rfc9724/, but that was achieved via the initials/surname route rather than asciiFullname so would still work.)

Comment (jennifer-richards) [reply rpc/lifecycle/tests_metadata.py:46]: re: rfc9005, https://datatracker.ietf.org/doc/rfc9005/ currently only lists that author as `C. Li`. Datatracker does not currently display non-latin titlepage names, only the latin transliteration. That should perhaps be a future feature, but we don't need to change that now.

Comment (kesara) [reply rpc/lifecycle/tests_metadata.py:46]: Resolving this because this is an attempt to match what DT's `titlepage_name` shows.
I created https://github.com/ietf-tools/datatracker/issues/11113 as DT enhancement.

File: rpc/lifecycle/metadata.py (hunk 1 above)
Comment (kesara) [reply rpc/lifecycle/metadata.py:306]: Resolving this because this is an attempt to match what DT's `titlepage_name` shows.
I created https://github.com/ietf-tools/datatracker/issues/11113 as DT enhancement.

**Review by kesara — APPROVED** [review]
LGTM

File: rpc/lifecycle/tests_metadata.py [hunk 3]
```diff
@@ -0,0 +1,69 @@
...
+                {"fullname": "Diane Egawa", "asciiFullname": "Gouda"}
+            ),
+            "Gouda",
+            "asciiFullname has priority + single name",
+        )
+        self.assertEqual(
+            Metadata.extract_name_from_author_dict({"fullname": "∂iane Egawa"}),
+            "∂. Egawa",
+            "fullname with two names",
+        )
```
kesara, line 59: Add a test case for `<author fullname="Q Misell" role="editor" surname="Misell">`?

jennifer-richards: As discussed elsewhere, let's do that in separate work. Fixing the corner cases of the name extraction is expanding the scope - we should do it, but it's not an immediate blocker

kesara: Created #1373 to capture this case.

rpc/lifecycle/metadata.py
chadholland125-ship-it, line 21: Chad Holland

PR #1379, opened 2026-07-02
feat: add field final review start date in pubq payload

rudimatz:
add field in pubq payload, as needed for https://github.com/ietf-tools/queue/issues/92


rudimatz:
value is already populated for internal queue, but not exposed yet

jennifer-richards (approved):

PR #1381, opened 2026-07-02
feat: concat text in abstract sub-tags

### LINKED ISSUE #1374: metadata extraction / comparison mangles abstract when it contains <tt> or other elements
For Send Email (Announce Publication), please handle the abstract in an expected manner when it contains elements. For example, it may contain the elements listed below. 

Suggestion: The abstract in the mail should match the output within the text file.  (At a minimum, please add a warning to the user that the Abstract needs to be replaced in the body of the mail.)

Specifically (XML to TXT):
`<tt>` yields nothing.
`<em>` yields surrounding underscores. Example: `_foo_`
`<strong>` yields surrounding asterisks. Example: `*foo*`
`<sub>` yields underscore. Example: `x_i`
`<sup>` yields carat. Example: `x^32`

Frequency: Likely rare bc past usage was low: 13 RFCs used `<tt>` in the abstract; 1 RFC used `<sup>`. (Other elements not used in an abstract afaict.)

Example: for rfc10007
- clicked 'Send Email', then 'Announce Publication'
- abstract was missing a bunch of text, seemingly due to presence of `<tt>` in `<abstract>` in the [XML file](https://github.com/rfc-editor-drafts/rfc10007/blob/main/rfc10007.xml). Screenshot below. 
- Workaround: copy & paste the abstract from the text file into the mail. and remove left margin. (Here's the [announcement](https://mailarchive.ietf.org/arch/msg/ietf-announce/IMjXppUuynM9EIerK_jsBz5weV8/) after manual correction.)

Image

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/1374 which broke the abstract string at the first occurrence of tags inside `<t>`

### DIFF
--- a/rpc/lifecycle/metadata.py
+++ b/rpc/lifecycle/metadata.py
@@ -21,15 +21,34 @@ from rpc.models import (
 logger = logging.getLogger(__name__)
 
 
+def _already_parenthesized(s: str) -> bool:
+    if len(s) < 2 or s[0] != "(" or s[-1] != ")":
+        return False
+    count = 0
+    for c in s[1:-1]:
+        count += 1 if c == "(" else -1 if c == ")" else 0
+        if count < 0:
+            return False
+    return count == 0
+
+
 def _is_simple_expression(expr: str) -> bool:
-    """Match xml2rfc's is_simple_expression: no parens needed for sub/sup."""
+    """Return True if this expression is simple enough to render without added
+    parentheses.
+
+    Accepts a single alphanumeric string with no whitespace, optionally preceded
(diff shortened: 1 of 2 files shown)


rudimatz:
use same logic as xml2rfc regarding markup tags like
```
<sub>, <sup>, <strong>, <em>
```

code reference https://github.com/ietf-tools/xml2rfc/blob/f448bd2da7409dfab915de9bb2a184604f0aa829/xml2rfc/writers/text.py#L3670

jennifer-richards (approved):

PR #1382, opened 2026-07-02
feat: display abstract, improve design of doc info tab

### LINKED ISSUE #1375: UI to allow updating abstract
Fixing #1374 will prevent future RFCs from having mangled abstracts pushed to datatracker, but will not resolve any existing data. We can perhaps fix the impact on the RFC that surfaced the issue, RFC 10007, through the admin, but it would be convenient if the RPC could correct these errors and sync them without admin intervention.

Hopefully a rare issue, though, so probably not a high priority if we can manually clean this up now.

rudimatz:
resolves https://github.com/ietf-tools/purple/issues/1375

image

### DIFF
--- a/client/app/components/DescriptionList.vue
+++ b/client/app/components/DescriptionList.vue
@@ -5,7 +5,7 @@ Based on https://tailwindui.com/components/application-ui/data-display/descripti
 -->
 <template>
   <div class="mt-3">
-    <dl class="divide-y divide-gray-100 dark:divide-gray-800">
+    <dl class="grid grid-cols-2 gap-x-6">
       <slot/>
     </dl>
   </div>
--- a/client/app/components/DescriptionListItem.vue
+++ b/client/app/components/DescriptionListItem.vue
@@ -1,6 +1,6 @@
 <template>
   <div :class="[
-    'w-full',
+    'w-full border-b border-gray-100 dark:border-gray-800',
     props.spacing === 'large' && 'px-4 py-3 sm:gap-4 sm:px-0',
     props.spacing === 'small' && 'px-1 py-0.5 sm:gap-0.5 sm:px-0',
   ]">
--- a/client/app/components/DocInfoCard.vue
+++ b/client/app/components/DocInfoCard.vue
@@ -416,6 +416,15 @@
             </PatchRfcToBeField>
           </DescriptionListDetails>
         </DescriptionListItem>
+        <DescriptionListItem term="Abstract" :spacing="spacing">
+          <DescriptionListDetails>
+            <PatchRfcToBeField fieldName="_abstract" :is-read-only="props.isReadOnly"
+              :ui-mode="{ type: 'textbox', placeholder: 'Abstract', rows: 8, initialValue: rfcToBe._abstract ?? '' }"
+              :draft-name="rfcToBe.name ?? ''" :on-success="props.refresh">
+              <span class="whitespace-pre-wrap text-sm">{{ rfcToBe._abstract }}</span>
+            </PatchRfcToBeField>
+          </DescriptionListDetails>
+        </DescriptionListItem>
       </DescriptionList>
     </div>
   </BaseCard>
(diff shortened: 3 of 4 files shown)


rudimatz:
also makes it editable BEFORE publication

jennifer-richards:
IMO the line wrapping artifacts due to the narrow column are distracting enough that this needs a different way to display the contents. Brainstorming, I think it'd be better e.g. to open a wider display when a button is clicked.

rudimatz:
putting it to DRAFT while reviewing display options

rudimatz:
@holloway do you have a good idea how this could be presented?

holloway:
I'm still unclear about the file format of 'abstract'. I'm pretty sure that in the past when an api gave me this field it would sometimes be a string with `<p>` tags. Was that the database value, or api level formatting of plaintext linebreaks into p tags?

I guess my suggestions about ux would depend on that.

jennifer-richards:
> I'm still unclear about the file format of 'abstract'.

In purple, the abstract in the database is currently straight text, no markup. It generally (but maybe not always) has explicitly wrapped lines, often with leading whitespace. It'll ~~normally~~ often be 72-character lines, I think. Paragraph breaks are double-newlines. The API that gives you `<p>`-wrapped abstracts is cooking that format.

It's possible in the future we'd add markup (HTML or markdown seem likely, but it could be something else like storing the actual RFCXML tags). If we did that, I think the interface would work on the literal source - i.e., editing the HTML or markdown, not as WYSIWYG.

IMO the main need is to be wide enough to display 72 char wide (or a bit wider, just in case) text without gross artifacts.

Edit: checked and RFCs published after Purple et al went live are not hard-wrapped. These are long-line-per-paragraph with a blank line at a paragraph break.

holloway:
Ok a lot of code in Red assumes abstract is possibly HTML so I'll update it.

Regarding UX suggestions...

On this route we've got these 3 columns but there's not an equal amount of content in each column. Generally it looks like there's 2.5vh (viewport height) worth of content in the first column, 1vh worth of content in the middle column, and 0.5vh worth of content in the last column. Perhaps it's worth rejigging the column layout to make more space for this wider abstract content. I suggest putting 3rd column 'Other labels' content below the 2nd column content, deleting the 3rd column, and making the remaining 2 column widths be 2/3 and 1/3.

ie, if the abstract column has 2 columns worth of space it will avoid the weird wrapping.

rudimatz:
@holloway sounds like a plan with reasonable effort and much clearer outcome. I'll do that

rudimatz:
@jennifer-richards @holloway 
something like this?

image

(the misaligned white boxes for complexities only came because of my extreme zoom)

holloway:
@rudimatz yes, looks good imo. Perhaps we should group them somehow or revise the order but I guess that can happen later.

What's with the abstract's first line having content pulled to the left? That looks odd.

holloway (approved):

holloway:
>  Perhaps we should group them somehow or revise the order but I guess that can happen later.

Already regretting saying this, after seeing that 'revision' is so tall being adjacent to authors :smile:

rudimatz:
> Already regretting saying this, after seeing that 'revision' is so tall being adjacent to authors 😄

might do a swap to balance it out

rudimatz:
new updated screenshot 
image

rudimatz:
waiting for input/comments from @ajeanmahoney about new design proposed

jennifer-richards:
The update looks good to me, fwiw.

> What's with the abstract's first line having content pulled to the left? That looks odd.

@holloway I think this is how the data are. I-D abstracts in datatracker are extracted from the rendered (or submitted) plaintext version, so the abstract is wrapped. That often ends up as the abstract in purple's database. I'd guess some step along the way is running `strip()` on the content, removing the indentation on the first line.

ajeanmahoney:
Looks good, some feedback from the team:

- The Document Info subheading could be cut because the tab label ("Info") already covers it
- When there are many authors, there is lots of whitespace around the Title field. 

Don't know if the whitespace can be tightened up, though, because consistent placement of these fields is a good thing (that is, I don't suggest floating the fields in the leftmost column up to remove whitespace)

PR #1390, opened 2026-07-09
fix: always build purple client

rjsparks:


docker/scripts/build-purple-client.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
jennifer-richards, line 1: This script is largely redundant with the existing `update-rpcapi` script. That other script also generates the Python client used to interact with datatracker. Breaking that into separate scripts is not a bad choice, but it should be updated to reduce redundancy if we do that.

The `update-rpcapi` does handle the situation that a datatracker is not available and will still generate the purple API client even in that case.

docker/scripts/build-purple-client.sh
```diff
@@ -0,0 +1,55 @@
...
+    sed -i '1i // @ts-nocheck' client/app/purple_client/runtime.ts
+fi
+
+# Keep a copy of the exact schema the client was built from.
+mkdir -p client/app/purple_client
+cp purple_api.yaml client/app/purple_client/.purple_api.yaml
+
+# Sanity check: every operationId in the schema should have a matching method
+# in the client (schema is snake_case, client methods are camelCase). This is
+# what catches the "... is not a function" class of drift before runtime does.
```
jennifer-richards, line 40: I don't think this sanity check section is a particularly useful addition at this point. It's just checking that the code generator generated the code it was told to do, and we only reach it if the code generator ran successfully.

docker/scripts/build-purple-client.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+#
+# Regenerate the purple OpenAPI schema (purple_api.yaml) and the generated
+# TypeScript client (client/app/purple_client/) from the CURRENT Django code.
+#
+# The generated client is gitignored, so it does NOT survive a container
+# rebuild and must be rebuilt from the running code. This is also the recovery
+# path whenever the frontend reports "api.<something> is not a function": that
+# error means the generated client is out of sync with the backend. Safe to
+# re-run any time (e.g. after a rebuild or a branch switch).
```
jennifer-richards, line 10: I don't think this is accurate if you're mounting the workspace in the dev container. The generated client will be written into your working directory, where it will persist even though it's gitignored. Tearing down the container won't delete it.

The generated client would not be present on a clean checkout, though. The comment would also be true if you're using a workspace entirely inside a docker volume.

jennifer-richards (changes requested):
Comments (which I've mostly given you offline already) inline.

If this is indeed fixing the issue with startup in the dev container, I suspect it's by preventing the app-init.sh script from bombing out if the `sed` or `/bin/cp` commands (lines 50 and 53 in app-init.sh) fail.

rjsparks:
Closing this as OBE

Issue #1347, opened 2026-06-11
Ability to generate queue reports


ajeanmahoney:
(maybe this isn't a Purple feature, but a new function, and probably needs sub issues)

Generate automatic reports that can be displayed on queue.rfc-editor.org:

Number of documents submitted, could actually be edited (block for missing reference cleared), and published
   - per year
   - per quarter
   - per month

In table format (the following is an example from last year, where PGTE means "Pages Gone To EDIT", i.e., pages that could be edited because the document's missing references finally entered the queue, so in modern terms, entered "In Progress" state for the first time):
<img width="607" height="231" alt="Image" src="https://github.com/user-attachments/assets/9c7d6773-a25e-4c3a-bb4b-cd8a10206cd2" />

In column format:
<img width="586" height="343" alt="Image" src="https://github.com/user-attachments/assets/db39d985-9eaf-4500-b501-20159d7811c6" />

For custom reports, provide the ability

- to set a date range for the report
- select stream (IETF) or area (WIT)

rudimatz:
@ajeanmahoney is this issue still valid or can we consider it done with current stats section?

ajeanmahoney:
We can close this one out. There are other issues on stats open, and there may be other requests for presentation later.

PR #1391, opened 2026-07-11
feat: statistics graphics and tables for the queue

rjsparks:


client/app/components/GroupedStackedBars.vue
```diff
@@ -0,0 +1,203 @@
...
+      .attr('font-size', 10).attr('font-weight', 600)
+      .text(d.label)
+  })
+}
+
+function showTooltip (event: MouseEvent, label: string, stream: string, status: string, count: number) {
+  const rect = container.value?.getBoundingClientRect()
+  tooltip.visible = true
+  tooltip.x = event.clientX - (rect?.left ?? 0) + 12
+  tooltip.y = event.clientY - (rect?.top ?? 0) + 12
```
holloway: Does x/y need to include the window scroll position to correctly position the tooltip?

client/app/pages/stats/stream.vue
```diff
@@ -0,0 +1,237 @@
...
+  { value: StatsQueuePeriodEnum.Quarter, label: 'Quarters' },
+  { value: StatsQueuePeriodEnum.Year, label: 'Years' },
+  { value: StatsQueuePeriodEnum.Ietf, label: 'IETF meetings' }
+]
+
+const STREAM_LABELS: Record<string, string> = {
+  ietf: 'IETF',
+  'ietf-wg': 'IETF WG',
+  'ietf-ad': 'IETF AD-sponsored',
+  ise: 'ISE', irtf: 'IRTF', iab: 'IAB', editorial: 'Editorial'
```
holloway: nit: formatting

client/app/pages/stats/stream.vue
```diff
@@ -0,0 +1,237 @@
...
+
+// The backend always splits IETF into ietf-wg / ietf-ad; off (default) merges
+// them back into a single "ietf" bucket. Client-side only — no refetch.
+const splitIetf = ref(false)
+const mergeStream = (slug: string) =>
+  (!splitIetf.value && (slug === 'ietf-wg' || slug === 'ietf-ad')) ? 'ietf' : slug
+
+// Deferred query controls: pending* bind to the inputs, applied* drive the query.
+const pendingPeriod = ref<StatsQueuePeriodEnum>(StatsQueuePeriodEnum.Year)
+const pendingCount = ref(4)
```
holloway: magic number? suggest extracting as a const to make this code clearer

client/app/pages/docs/[id]/timeline.vue
```diff
@@ -0,0 +1,186 @@
...
+
+      <BaseCard class="w-full grid place-items-stretch">
+        <div class="flex items-center justify-between">
+          <h3 class="text-base font-semibold leading-7">
+            Assignment timeline
+            <Icon v-show="timelineStatus === 'pending'" name="ei:spinner-3" size="1.5em" class="animate-spin" />
+          </h3>
+          <div class="flex items-center gap-3 text-xs">
+            <span class="flex items-center gap-1">
+              <span class="inline-block h-3 w-3 rounded-sm" :style="{ backgroundColor: KIND_COLORS[KIND_WORKING] }" />
```
Comment (holloway) [inline client/app/pages/docs/[id]/timeline.vue]: should we use conditional tailwind classes for these instead of directly setting colors? i'm guessing this won't have darkmode variations eg,

```suggestion
              <span :class="['inline-block h-3 w-3 rounded-sm', KIND_CSS_CLASSES[KIND_WORKING] ]" />
```

client/app/components/StackedTimeBars.vue
```diff
@@ -0,0 +1,287 @@
...
+  isBlocked: boolean
+  isOther: boolean
+  members: string[] // role slugs summed into this category
+  color: string
+}
+
+const container = ref<HTMLElement | null>(null)
+const svgEl = ref<SVGSVGElement | null>(null)
+const width = ref(720)
+const height = 340
```
holloway: should the height be uppercase to indicate a constant?

does the width need to be recomputed when the browser resizes?

client/app/components/GroupedStackedBars.vue
```diff
@@ -0,0 +1,203 @@
...
+    <!-- Interactive legend: click a status to hide/show it (chart rescales). -->
+    <div v-if="statuses.length" class="mt-3 text-xs">
+      <p class="mb-1 opacity-50">Color = status; each group is one period, one bar per stream. Click a status to hide it.</p>
+      <div class="flex flex-wrap gap-x-4 gap-y-1">
+        <button
+          v-for="s in statuses" :key="s"
+          type="button"
+          class="flex items-center gap-1 rounded px-1 transition-opacity hover:bg-gray-100 dark:hover:bg-neutral-800"
+          :class="hidden.has(s) ? 'opacity-35' : ''"
+          @click="toggle(s)"
```
holloway:
```suggestion
  :aria-pressed="!hidden.has(s)"
```

client/app/components/GroupedStackedBars.vue
```diff
@@ -0,0 +1,203 @@
...
+        const yTop = y(cursor + c)
+        const h = Math.max(y(cursor) - yTop - SEG_GAP, 0)
+        if (h > 0) {
+          svg.append('rect')
+            .attr('x', sx).attr('y', yTop)
+            .attr('width', barW).attr('height', h)
+            .attr('rx', 1)
+            .attr('fill', statusColor(status))
+            .style('cursor', 'pointer')
+            .on('mousemove', (e: MouseEvent) => showTooltip(e, d.label, stream, status, c))
```
holloway, line 178: also support keyboard focus-based tooltips?

client/app/utils/statsViz.ts
```diff
@@ -0,0 +1,119 @@
+import humanizeDuration from 'humanize-duration'
+import type { TimelineSegment } from '~/purple_client'
+
+// Segment kinds emitted by the backend (rpc.lifecycle.timeline).
+export const KIND_BLOCKED = 'blocked'
+export const KIND_WORKING = 'working'
+export const KIND_LEGACY = 'legacy_label'
+export const KIND_AWAITING = 'awaiting_ref'
```
holloway: Can we expose these via the client or should we hardcode a copy here?

Separate issue, considering these are used to index things like `KIND_COLORS` below a narrow type might be useful eg

```typescript
export type Kind = typeof KIND_BLOCKING | typeof KIND_WORKING | typeof KIND_LEGACY | typeof KIND_AWAITING
```

and then you could use that index in `KIND_COLORS` below like, 

```
KIND_COLORS: Record<Kind, string> =
```

of course there's also the other comment about tailwind classes being preferable so this is just a general idea to narrow types where possible, rather than making a record index  `string`.

client/app/pages/stats/counts.vue
```diff
@@ -0,0 +1,192 @@
...
+
+const api = useApi()
+
+const PERIOD_OPTIONS = [
+  { value: StatsQueuePeriodEnum.Week, label: 'Weeks' },
+  { value: StatsQueuePeriodEnum.Month, label: 'Months' },
+  { value: StatsQueuePeriodEnum.Quarter, label: 'Quarters' },
+  { value: StatsQueuePeriodEnum.Year, label: 'Years' },
+  { value: StatsQueuePeriodEnum.Ietf, label: 'IETF meetings' }
+]
```
holloway, line 158: This seems to be repeated for several routes. Move to a utils file?

client/app/pages/stats/counts.vue
```diff
@@ -0,0 +1,192 @@
...
+
+// Query controls are deferred (the rollup can run long): pending* bind to the
+// inputs, applied* drive the query and only change on Apply.
+const pendingPeriod = ref<StatsQueuePeriodEnum>(StatsQueuePeriodEnum.Month)
+const pendingCount = ref(6)
+const appliedPeriod = ref<StatsQueuePeriodEnum>(pendingPeriod.value)
+const appliedCount = ref(clamp(pendingCount.value))
+
+function clamp (n: number): number {
```
holloway: There's a lodash `clamp` util that can do this.

client/app/pages/stats/stream.vue
```diff
@@ -0,0 +1,237 @@
...
+  return statuses.value.reduce((sum, s) => sum + count(p, stream, s), 0)
+}
+function periodTotal (p: QueuePublishedStatPeriod): number {
+  return p.counts.reduce((sum, c) => sum + c.count, 0)
+}
+const fmt = (n: number) => (n === 0 ? '—' : n.toLocaleString())
+
+const showWeekRange = computed(() => /^\d{4}-W\d{2}$/.test(periods.value[0]?.label ?? ''))
+function weekRange (p: QueuePublishedStatPeriod): string {
+  const f = (d: Date) => d.toLocaleDateString('en-US', { timeZone: 'UTC', month: 'short', day: 'numeric' })
```
holloway: afaik we should be using Luxon dates for this kind of stuff not the native `Date`

client/app/pages/stats/index.vue
```diff
@@ -0,0 +1,278 @@
...
+  }
+)
+
+const periods = computed(() => stats.value?.periods ?? [])
+
+// Week labels (2026-W28) are terse; show the covered date range beneath them.
+// Keyed off the displayed data, not appliedPeriod, so stale rows during a
+// refetch keep their correct sublabels.
+const showWeekRange = computed(() => /^\d{4}-W\d{2}$/.test(periods.value[0]?.label ?? ''))
+function weekRange (p: QueuePeriodStat): string {
```
holloway: Some more things that should probably be deduplicated (`showWeekRange` regex and `weekRange` itself)

client/app/utils/statsViz.ts
```diff
@@ -0,0 +1,119 @@
...
+export function humanSeconds (seconds: number): string {
+  return humanMillis(seconds * 1000)
+}
+
+/** Whole-day rendering for the queue summary (never sub-day units). */
+export function formatDays (seconds: number): string {
+  if (seconds <= 0) {
+    return '—'
+  }
+  const days = Math.round(seconds / 86400)
```
holloway: nit: extract `86400` as a named constant.. bonus points for using `_` numeric separators

client/app/components/StackedTimeBars.vue
```diff
@@ -0,0 +1,287 @@
...
+
+const MARGIN = { top: 16, right: 16, bottom: 52, left: 64 }
+
+function weekRange (p: QueuePeriodStat): string {
+  const fmt = (d: Date) =>
+    d.toLocaleDateString('en-US', { timeZone: 'UTC', month: 'short', day: 'numeric' })
+  const last = new Date(p.end.getTime() - 86400000) // end is exclusive (next Monday)
```
holloway: nit: `86400000` same thing

client/app/components/StackedTimeBars.vue
```diff
@@ -0,0 +1,287 @@
...
+  return `${fmt(p.start)} – ${fmt(last)}`
+}
+const SEG_GAP = 2 // px of surface between stacked segments (dataviz mark spec)
```
holloway: a very subjective take but add `_PX` suffix (and `_MS` suffix, elsewhere) units to numbers

rudimatz (approved):
probably a good moment to move all "tests" files into separate `tests` folder

jennifer-richards:
> probably a good moment to move all "tests" files into separate `tests` folder

I don't exactly object, but that is a change from the structure we've used in datatracker and elsewhere. It'll require some superficial changes (particularly for relative imports, with `.` -> `..`) and a `tests/__init__.py` file.

My own preference is just to leave them alongside the modules they test, but that's just one vote.

rudimatz:
mentioning here https://github.com/ietf-tools/purple/issues/1347, TBD if this PR relates to that (can logic be re-used?)

holloway:
I'm in favor of keeping tests colocated with their code

client/app/components/GroupedStackedBars.vue
rjsparks: resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9

client/app/pages/stats/stream.vue
rjsparks: resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9

client/app/pages/stats/stream.vue
rjsparks: resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9

client/app/components/StackedTimeBars.vue
rjsparks: resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9

client/app/components/GroupedStackedBars.vue
rjsparks: resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9

client/app/utils/statsViz.ts
rjsparks: partially resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9. As of that commit they're still in the frontend instead of being provided by the backend,

client/app/pages/stats/counts.vue
rjsparks: resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9

client/app/pages/stats/counts.vue
rjsparks: resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9

client/app/pages/stats/stream.vue
rjsparks: resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9

client/app/pages/stats/index.vue
rjsparks: resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9

client/app/utils/statsViz.ts
rjsparks: resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9

client/app/components/StackedTimeBars.vue
rjsparks: resolved in 4ea57d6a0e3cd90b93dcc241ea53d1643eedc9c9

rjsparks:
> mentioning here https://github.com/ietf-tools/purple/issues/1347, TBD if this PR relates to that (can logic be re-used?)

Yes. We'll iterate on this with the RPC for a little bit, then reuse much of this for views on the queue site.

client/app/components/GroupedStackedBars.vue
rjsparks: I would prefer to leave this for a focused followup after this PR.

client/app/components/StackedTimeBars.vue
rjsparks: addressed in 964f972b55a34d5b463ea2923f92be4413a7735e

client/app/utils/statsViz.ts
rjsparks: backend enum part of the request resolved in fa7e727109e38843eeb7b50df6b4e719a78f838b

rjsparks:
@holloway - d01df98df1fb176e587c5b8b1ab409d15d318055 deserves focused inspection

holloway (approved):

rpc/stats/intervals.py
```diff
@@ -0,0 +1,185 @@
+# Copyright The IETF Trust 2026, All Rights Reserved
```
jennifer-richards, line 1: Probably should move the new code (in this and other files) into a separate package - add `stats` alongside `lifecycle` to keep the latter from becoming a dumping ground for random code.

This module is strange -- it's _mostly_ generic "pure interval algebra" code as advertised, but then mixes in a little bit of Assignment-specific details for dealing with states. It then marks _all_ of its methods as non-public with `_` prefixes. That's a valid naming convention, except that it's really not clear where the boundary of that API is meant to be.

That might become clearer by moving this into the `stats` package, I guess.

rpc/stats/timeline.py
```diff
@@ -0,0 +1,407 @@
...
+KIND_BLOCKED = "blocked"
+
+
+KIND_WORKING = "working"
+
+
+KIND_LEGACY = "legacy_label"
+
+
+KIND_AWAITING = "awaiting_ref"
```
jennifer-richards, line 70: style nit: odd blank lines

rpc/tests_intervals.py
```diff
@@ -0,0 +1,96 @@
+# Copyright The IETF Trust 2026, All Rights Reserved
```
jennifer-richards: This file should be in the same directory as intervals.py

The set of things being tested seems arbitrary. Many of the methods in intervals.py don't have explicit tests. Maybe that's because they're tested implicitly, I suppose, but that needs explanation. 

E.g., `_split_runs_by_intervals()` is only ever used in `timeline.py` and has no test.

rpc/stats/tests_rollups.py
```diff
@@ -0,0 +1,593 @@
+# Copyright The IETF Trust 2026, All Rights Reserved
```
jennifer-richards, line 1: This file should be in the same directory as rollups.py

rpc/stats/tests_timeline.py
```diff
@@ -0,0 +1,248 @@
+# Copyright The IETF Trust 2026, All Rights Reserved
```
jennifer-richards, line 1: This file should be in the same directory as timeline.py

rpc/stats/test_helpers.py
```diff
@@ -0,0 +1,104 @@
+# Copyright The IETF Trust 2026, All Rights Reserved
```
jennifer-richards, line 1: Belongs in the directory with the modules that use it, and should probably just be `test_helpers.py`.

rpc/lifecycle_test_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+import datetime
+
+from .factories import AssignmentFactory, RfcToBeFactory, RpcRoleFactory
+from .models import DocRelationshipName, RpcRelatedDocument
+
+UTC = datetime.UTC
+DAY = datetime.timedelta(days=1)
+
+
+def _dt(year, month, day):
```
jennifer-richards: Eww. If we actually want this helper, it should be something we use project-wide.

rpc/lifecycle/rollups.py
```diff
@@ -0,0 +1,920 @@
...
+_HistRec = namedtuple("_HistRec", ["state", "history_date"])
+
+
+def _add_months(year: int, month: int, delta: int) -> tuple[int, int]:
+    """Return ``(year, month)`` shifted by ``delta`` months (month is 1-12)."""
+    index = (year * 12 + (month - 1)) + delta
+    return index // 12, index % 12 + 1
+
+
+def _aware(d: datetime.date) -> datetime.datetime:
```
jennifer-richards: This helper's name is concerning—code should be very explicit about turning a date into midnight UTC on that date.

rpc/lifecycle/rollups.py
```diff
@@ -0,0 +1,920 @@
...
+
+def period_windows(period: str, count: int, now: datetime.datetime) -> list[dict]:
+    """Windows, oldest first, ending with the current period.
+
+    ``period`` is one of ``week`` / ``month`` / ``quarter`` / ``year`` (all
+    calendar-aligned) or ``ietf`` (spans between consecutive IETF meetings).
+    """
+    if count < 1:
+        return []
+    today = now.date()
```
jennifer-richards: This is a timezone-dependent operation. It'd be best to force it to UTC before taking `.date()`, particularly because the `_aware()` method then assumes UTC to convert back to a datetime

jennifer-richards (changes requested):
Some comments on the python code are inline. Overall, it seems to be functional, but I'm concerned about the structure and maintainability of the new code. It needs to be organized into a package instead of piled into the `lifecycle` package. The tests look like they may exercise every line, but it's very difficult to see why particular methods have tests and others don't. In general, it's hard to tell what is a one-off helper method and what is a possibly reusable piece of code. I think a pass at cleaning up that structure is needed before merging.

rjsparks:
☝️ that flurry is a rebase to the oxc clean main.

rjsparks:
The last three commits are addressing Jennifer's review, and making the new code oxc clean.

jennifer-richards (approved):
Still has the `dt()` helper. It's useful enough here to keep, but I think we should consider a refactor to make that into a general utility for the entire project - i.e., an equivalent to `datetime.datetime(..., tzinfo=datetime.UTC)`, not just a special-case for dates. That should be a separate PR though.

Thing to watch is that the `tests_intervals.py` file uses `SimpleTestCase`. That's appropriate because the tests don't hit the database, but is a potential footgun in future work.

rudimatz:
@rjsparks I think you can just run a "./update-rpcapi" (followed by the oxfmt) and that should fix the errors. API client seems to be stale

PR #1393, opened 2026-07-13
feat: add SimpleHistory for untracked models

### LINKED ISSUE #1308: Verify SimpleHistory usage on models
its missing on `ActionHolder`. check which models should have it additionally

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1308

add SimpleHistory for previously untracked models:
- datatracker.Documents
- ActionHolder
- RpcAuthorComment
- RpcPerson

### DIFF
--- a/datatracker/migrations/0003_historicaldocument_historicaldocumentlabel.py
+++ b/datatracker/migrations/0003_historicaldocument_historicaldocumentlabel.py
@@ -1,0 +1,124 @@
+# Copyright The IETF Trust 2026, All Rights Reserved
+
+import django.db.models.deletion
+import simple_history.models
+from django.conf import settings
+from django.db import migrations, models
+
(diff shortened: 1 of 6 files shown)


jennifer-richards (commented):
(just to note: this is still draft, so making a comment rather than approval)

Seems ok. Do we still have a hook that creates initial history when we add history to the model? I don't remember offhand if that is in our startup scripts or if it was part of the xfer process.

Are we going to need to backfill the history for any of these?

rudimatz:
added trigger for QUEUE notification on ActionHolder and display it in history tab

rudimatz:
> Are we going to need to backfill the history for any of these?

no, no need for backfill.

jennifer-richards (approved):

PR #1396, opened 2026-07-15
feat: add modal to create new assignment

### LINKED ISSUE #1334: Add ability to assign a document from the Assignments & Authors tab
Currently, the user has to go to the main queue page to make an assignment, but frequently they are looking at the document's page (e.g., seeing if it's a good fit for an editor or setting up the final review and needing to assign the final review editor). It would be convenient to have the ability to assign the document from the document's Assignments & Authors tab. The button can be placed next to the pending activity label like it is on the main queue page:

Image

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1334

### DIFF
--- a/client/app/components/AssignmentFinishedModal.vue
+++ b/client/app/components/AssignmentFinishedModal.vue
@@ -7,9 +7,17 @@
           {{ props.rfcToBe.name }}
         </span>
       </h1>
-      <BaseButton btnType="cancel" class="m-2 flex items-center" @click="closeOverlayModal">
-        <Icon name="uil:times" class="h-5 w-5" aria-hidden="true" />
-      </BaseButton>
+      <div class="m-2 flex items-center gap-2">
+        <BaseButton btnType="default" class="flex items-center" @click="openAddAssignmentModal">
+          <span>Add assignment</span>
+          <span v-if="isLoadingAdd" class="w-3">
+            <Icon name="ei:spinner-3" size="1rem" class="animate-spin" />
+          </span>
+        </BaseButton>
+        <BaseButton btnType="cancel" class="flex items-center" @click="closeOverlayModal">
+          <Icon name="uil:times" class="h-5 w-5" aria-hidden="true" />
+        </BaseButton>
+      </div>
     </div>
     <div class="flex-1 overflow-y-scroll px-4 pt-4 pb-7">
       <ul class="flex flex-col gap-4">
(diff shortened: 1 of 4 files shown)


client/app/components/AssignmentFinishedModal.vue
```diff
@@ -72,5 +99,59 @@ if (!overlayModalKeyInjection) {
...
+    // Default to the draft's next pending activity (earliest in pipeline order),
+    // falling back to the first selectable role.
+    const pending = [...(props.rfcToBe.pendingActivities ?? [])].sort(
+      (a, b) => roleOrderIndex(a.slug) - roleOrderIndex(b.slug)
+    )
+    const defaultRole = pending[0]?.slug ?? roleOptions[0]?.slug ?? ''
+
+    const peopleWorkload = calculatePeopleWorkload(clusters, queueItems)
+
+    openOverlayModal({
```
holloway, line 139: `await` this?

It seems we have inconsistent patterns across the codebase for this so I'm not sure what the correct pattern is.

If it is awaited then the noop `.catch(() => {})`  is probably redundant, considering the following catch.

client/app/components/AssignmentModal.vue
```diff
@@ -229,12 +261,12 @@ watch(
             }
           })
       )
-    } else if (props.message.type === 'assign') {
-      const message = props.message
+    } else if (props.message.type === 'assign' || props.message.type === 'add') {
+      const rfcToBeId = props.message.rfcToBeId
```
holloway, line 262: nit
```suggestion
      const { rfcToBeId } = props.message
```

holloway (approved):
lgtm only minor feedback

client/app/components/AssignmentFinishedModal.vue
jennifer-richards: Since it's in an async, awaiting would make sense to me. . .

jennifer-richards (commented):
Tying the modal's ability to select the role / hide the blocked role specifically to the `add` message feels awkward to me. I'd have expected that to be an explicit option to the modal instead of tying its functionality that tightly to the workflows we have at the moment.

Just a comment because I mostly defer to @holloway on the PR overall.

jennifer-richards (approved):
Refactor looks good to me. Not sure if you wanted to adopt or reject the suggestion to use `await` though.

PR #1405, opened 2026-07-18
feat: oxc format API client after generation

rudimatz:
relates to https://github.com/ietf-tools/purple/pull/1404


rudimatz:
using "npm ci" in the app-init script to prevent regeneration of lockfile (which produced changes in local development)

jennifer-richards (approved):

PR #1407, opened 2026-07-20
fix: resolve type-aware oxlint errors surfaced by oxlint-tsgolint

rjsparks:
Enabling oxlint's type-aware rules (via the oxlint-tsgolint engine) surfaced 13 pre-existing errors across 8 files. None are behavioural except the first:

- FinalReviews: the queue filter callback used a block body but never returned, so `queueItem.name == props.name` was dead and the filter dropped every item; return the (strict) comparison instead
- drop unused imports/bindings: RpcPerson (url), RfcToBe (queue), the catch binding (snackbar), and mark colorMode unused (document_relations)
- use String(...) instead of template literals on unknown values (snackbar)
- new Set(x) instead of new Set([...x]) — no-useless-spread (docs pages)
- drop redundant type constituents: `Relationship | string` -> `string`, and the explicit `undefined` on an optional param (document_relations)

Co-Authored-By: Claude

### DIFF
--- a/client/app/components/FinalReviews.vue
+++ b/client/app/components/FinalReviews.vue
@@ -49,9 +49,7 @@ const queueListWithFilters = computed(() => {
   if (!props.name) {
     return queueList.value
   }
-  return queueList.value.filter((queueItem) => {
-    queueItem.name == props.name
-  })
+  return queueList.value.filter((queueItem) => queueItem.name === props.name)
 })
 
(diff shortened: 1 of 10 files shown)


NGPixel (approved):

rudimatz:
@NGPixel  can you look at (and approve) my last commit high jacking this PR? This should avoid the ongoing issues with "commander" that previously made the tests fail on this PR

NGPixel (approved):

PR #1419, opened 2026-07-21
feat: remove traces of publication_stream

rudimatz:
as discussed, `publication_stream` shall be removed from project to avoid confusion

### DIFF
--- a/client/app/components/DocInfoCard.vue
+++ b/client/app/components/DocInfoCard.vue
@@ -177,14 +177,6 @@
               :on-success="props.refresh">
               <span class="flex-1">
                 {{ rfcToBe.stream?.name ?? rfcToBe.stream?.slug }}
-                <span
-                  v-if="
-                    rfcToBe.publicationStream &&
-                    rfcToBe.publicationStream.slug !== rfcToBe.stream?.slug
-                  ">
-                  (published as
-                  {{ rfcToBe.publicationStream.name ?? rfcToBe.publicationStream.slug }})
-                </span>
               </span>
             </PatchRfcToBeField>
           </DescriptionListDetails>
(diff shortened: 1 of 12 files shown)


rudimatz:
@alicerusso @ajeanmahoney I just wanna confirm that its still valid as we discussed in Vienna, we don't need the `publication_stream` field and want to remove it, right?

alicerusso:
Yes, that's my understanding. To summarize, it seems the field was intended to hold an RFC's "stream at time of publication" but it's holding odd data for 2400+ records.  Specifically, here's the note from Slack re: the 2460 records where publication_stream_id="ise" and stream_id != "ise":
- To be decided: Either drop the publication_stream_id column or stop displaying "(published as foo)" for stream in Purple.
- For background: "ise" was used as the fallback value for publication_stream_id at data xfer time.

rudimatz:
FYI @rjsparks

jennifer-richards (commented):
Does this have any risk of corrupting the history records? It drops the field from the historical models. If there were any change records that only changed the `publication_stream` field, those will be false changes. Do we have such entries, and if so is that safe or does it need further cleanup?

rudimatz:
good point to pay extra attention on this, but I tested locally, behavior is as expected, no wrong history entries

jennifer-richards:
Discussed offline with Rudi. In theory the migration could leave behind history records that indicate there was an update but with no actual change: this would happen if the only field that had changed is the `publication_stream`. This would not be an issue, but could possibly make the history look funny if it reported a change without a change. Rudi checked the existing histories and did not find any cases where this would happen, though, so it's moot.

jennifer-richards (approved):

PR #1421, opened 2026-07-22
feat: provide names (not only slug) for UI

### LINKED ISSUE #1417: for Stream, change "ise" to "Independent"
from any /docs/ page, if you click "Stream", the option in the dropdown is "ise". should be "Independent".

rationale: accurate stream name.

(and it should be displayed as Independent. for example: on /docs/rfc9932 "Stream ise" should be "Stream Independent")

### LINKED ISSUE #1430: Info tab: Show full Status description
Instead of the abbreviation for Status (e.g., "ps"), show the full text of the status ("Proposed Standard").

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/1417 and fix https://github.com/ietf-tools/purple/issues/1430
display in UI the NAME instead of the slug, for disposition, relationships, std_level and stream.
also expose these as ADDITIONAL fields in pubq API

needs additionally a DB change for stream name -> DONE

### DIFF
--- a/rpc/serializers.py
+++ b/rpc/serializers.py
@@ -915,6 +915,29 @@ class RfcToBeSerializer(serializers.ModelSerializer):
         source="additionalemail_set", many=True, read_only=True
     )
     blocking_reasons = RfcToBeBlockingReasonSerializer(many=True, read_only=True)
+    # Human-readable names for the slug-valued FK fields, for display. The plain
+    # slug fields stay writable for editing.
+    disposition_name = serializers.SlugRelatedField(
+        source="disposition", slug_field="name", read_only=True
+    )
+    stream_name = serializers.SlugRelatedField(
+        source="stream", slug_field="name", read_only=True
+    )
+    publication_stream_name = serializers.SlugRelatedField(
+        source="publication_stream", slug_field="name", read_only=True
+    )
+    std_level_name = serializers.SlugRelatedField(
+        source="std_level", slug_field="name", read_only=True
+    )
+    publication_std_level_name = serializers.SlugRelatedField(
+        source="publication_std_level", slug_field="name", read_only=True
+    )
+    boilerplate_name = serializers.SlugRelatedField(
+        source="boilerplate", slug_field="name", read_only=True
+    )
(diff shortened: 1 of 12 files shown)


jennifer-richards (commented):
Just commenting to capture our offline discussion. I suggest investigating whether passing a structured field for the affected fields would be better. E.g., the equivalent of `{"id": 3, "slug": "myslug", "name": "My Slug Is Friendly"}`. (The "id" field might not be necessary,.)

rudimatz:
following recent comment, I use nested obj using NameSerializer. Like this, UI can display names and use slugs where needed.
affected fields:

- disposition
- stream
- publication_stream
- std_level
- publication_std_level
- boilerplate
- submitted_format

rpc/serializers.py
```diff
@@ -2565,6 +2640,9 @@ class Meta:
             "authors",
             "approval_log_message",
             "stream",
+            "stream_name",
+            "std_level_name",
+            "disposition_name",
```
jennifer-richards, line 2649: Why the new `_name` fields instead of changing the serializers for the existing stream/std_level/disposition fields? (In other places, the new fields were `_slug` for writing, what's different here?)

jennifer-richards (commented):
Question inline

Also an observation: the `NameSerializer`-powered fields lose their documentation strings in the schema. That's not a problem here, but it might be worth seeing if there's a way to preserve those while using the shared serializer. If this pattern proves useful we might want to use it somewhere that the API documentation is more important (i.e., in a less private API)

rpc/serializers.py
rudimatz: I kept those to be backwards compatible. This is `PublicQueueItemSerializer` used by external queue page.
I agree its not 100% consistent, but these are separate APIs used by independent clients, I think its reasonable to not break the schema (instead of doing a concerted release of multiple projects required with a breaking change)

rudimatz:
@jennifer-richards I re-added the lost descriptions for the API schema

jennifer-richards (approved):

PR #1431, opened 2026-07-23
perf: eliminate N+1 queries in the queue list (426→9)

rjsparks:
Rendering GET /api/rpc/queue/ ran 426 SQL queries for 154 in-queue documents, growing linearly with the queue. Three per-item lookups in QueueItemSerializer were responsible:

- RfcToBe.cluster calls draft.cluster_set.first() per row
- pending_activities() filters assignment_set per row
- pending_activities serialized RpcRole.objects.filter(...) per row

Batch all three:

- add RfcToBeQuerySet.with_cluster(), prefetching draft__cluster_set with an explicit ordering so cluster_set.first() is served from the prefetch cache (an unordered prefetched queryset's .first() clones and re-queries)
- add RfcToBeQuerySet.with_activity_assignments(), prefetching the non-withdrawn/closed assignments to activity_assignments; pending_activities() uses them when present and falls back to the query otherwise
- make QueueItemSerializer.pending_activities a SerializerMethodField that loads RpcRole once per request (the child serializer is shared for many=True) rather than once per row; schema preserved via extend_schema_field

QueueList chains the new methods; PublicQueueList and QueueCounts are unaffected. The queue now issues a constant 9 queries regardless of size. Output of QueueItemSerializer and PublicQueueItemSerializer is byte-for-byte identical before and after (verified against the dev DB); full rpc suite passes.

Co-Authored-By: Claude


rudimatz (approved):

rudimatz:
relates to https://github.com/ietf-tools/purple/issues/222

PR #1432, opened 2026-07-23
feat: add is_public flag to labels

### LINKED ISSUE #668: Allow user to indicate whether label should be shown on public queue page
When creating or editing a label. For example, a check box that says, "Display label on public queue page"

rudimatz:
resolves https://github.com/ietf-tools/purple/issues/668


rudimatz:
also remove non-public labels from pubq endpoint (so they don't appear on queue page)

jennifer-richards (approved):

PR #1447, opened 2026-08-10
feat: migration to backfill rev history

### LINKED ISSUE #1433: History displays a "rev" change that is incorrect
with https://github.com/ietf-tools/purple/blob/main/rpc/migrations/0010_add_rfctobe_rev.py we backfilled the "rev" value for newly added field `rctobe.rev`

However this `save()` didn't trigger a capture in history, so that now the rev change will be attributed to the **next** person that does a change.

e.g. https://purple.rfc-editor.org/docs/draft-nerenberg-imap-binary/history shows incorrectly that Alice set the rev, in addition to changing the "stream" which actually happened.

rudimatz:
fixes https://github.com/ietf-tools/purple/issues/1433

adds the current `rev` value into history so that it will not be falsely attributed to the the next editor that does a change


rudimatz:
this script will also clean up EXISTING wrong history entries for `rev` that are misattributed, like 
image

rudimatz:
with latest commit, `rev` changes made by humans will be preserved

rpc/migrations/0014_backfill_rev_into_history.py
```diff
@@ -0,0 +1,35 @@
...
+    """Fill the rev that 0010 set into the snapshots that never recorded it.
+
+    0010 set RfcToBe.rev without creating history (migrations bypass
+    simple-history), so the value would otherwise surface as a change on
+    the next editor's save. Only the empty ('') snapshots — the ones predating
+    the field being populated — are filled, with the earliest rev history did
+    record; genuine rev values are left untouched.
+    """
+    RfcToBe = apps.get_model("rpc", "RfcToBe")
+    Historical = apps.get_model("rpc", "HistoricalRfcToBe")
```
jennifer-richards: nit: suggest naming it `HistoricalRfcToBe` by convention

rpc/migrations/0014_backfill_rev_into_history.py
```diff
@@ -0,0 +1,35 @@
...
+    for rfctobe in RfcToBe.objects.exclude(rev=""):
+        snaps = Historical.objects.filter(id=rfctobe.id)
+        first_real = (
+            snaps.exclude(rev="")
+            .order_by("history_date", "history_id")
+            .values_list("rev", flat=True)
+            .first()
+        ) or rfctobe.rev
+        snaps.filter(rev="").update(rev=first_real)
```
jennifer-richards, line 25: I'm pretty sure the answer is "yes," but are we certain that every existing record where the rev is blank should get the `first_real` value? I.e., we don't have strange situations where it was set, updated, then cleared?

What I am worrying about is a case where the rev changed between history records in a sequence like `[None, None, 1, 2, None]`. This should become `[1, 1, 1, 2, 2]` and not `[1, 1, 1, 2, 1]`. But, like I said, this is not very likely - just want to confirm.

jennifer-richards (changes requested):
One nit to change and a highly paranoid question about assumptions.

(Since this is just a migration I hesitated to ask to fix the nit but I think we should if only to make searches on `HistoricalRfcToBe.objects` find all the places we've touched that)

rpc/migrations/0014_backfill_rev_into_history.py
rudimatz: I validated, that case doesn't exist

jennifer-richards (approved):

PR #1454, opened 2026-08-11
feat: show shepherd name instead DT id

### LINKED ISSUE #1436: Ingest Document page: for Doc Shepherd, it's displaying a number instead of a name
example 1: https://purple.rfc-editor.org/docs/import/?documentId=104441
shows "Document Shepherd: 125623"
screenshot below.

example 2: https://purple.rfc-editor.org/docs/import/?documentId=136502
shows "Document shepherd: 140327"

(These examples are from clicking the doc-specific links on https://purple.rfc-editor.org/queue/submissions
They will no longer be available as the documents are enqueued.)

Image

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1436

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -489,6 +489,11 @@ def submission(request, document_id, rpcapi: rpcapi_client.PurpleApi):
     if draft is None:
         raise NotFound(f"No draft found with id {document_id}")
     subm = Submission.from_rpcapi_draft(draft)
+    # draft.shepherd is a datatracker person id; resolve it to a display name.
+    if draft.shepherd is not None:
+        with datatracker_api():
+            shepherd = rpcapi.get_person_by_id(draft.shepherd)
+        subm.shepherd = shepherd.plain_name if shepherd is not None else ""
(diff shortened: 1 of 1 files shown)


rudimatz:
image
see issue

jennifer-richards (approved):
Might want to handle exceptions if it's not already handled outside of this. If the shepherd lookup happens to fail, you could show an error in that field but otherwise work. Not sure if that makes sense in context though, so just a thought.

rudimatz:
@jennifer-richards though a mismatch should not ever happen (DT provides shepherd ID, and then gives name of same shepherd ID), I will add a message in UI if shepherd could not be resolved

jennifer-richards:
> @jennifer-richards though a mismatch should not ever happen (DT provides shepherd ID, and then gives name of same shepherd ID), I will add a message in UI if shepherd could not be resolved

I was really thinking about datatracker entirely failing to respond  to the second API call for some reason.

jennifer-richards (approved):

PR #1457, opened 2026-08-13
feat: move activity when all previous steps done

### LINKED ISSUE #1406: The doc state shouldn't change until all editors have completed their assignments
When an editor provides assistance on a document (e.g., helps another editor with sourcecode validation) and then completes their assignment by providing their hours, the document should not change state until the other editor provides their hours. That is, state shouldn't change until all editors assigned to a state (e.g., Final Review) enter their hours.

From one of the editors, who handled Final Review while the main editor was OOO:
> I assigned myself as a second Final Review editor for Kaelin’s doc. But when I passed the doc back to her and added my hours for Final Review, the public queue showed both “In Final Review” and “Awaiting Publisher”. The internal queue showed Kaelin as current Final Review Editor and showed the doc as ready to assign to a publisher.

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1406

if multiple editors work on doc on same assignment role, only consider it done (and show next pending activity), if ALL those assignments are done

also adding tests

### DIFF
--- a/rpc/lifecycle/activities.py
+++ b/rpc/lifecycle/activities.py
@@ -21,14 +21,14 @@ class CompletedAssignment(Activity):
             self.prereqs = prereqs
 
 
+# Formatting and reference checking run in parallel once the doc is enqueued;
+# first edit waits for both.
 ENQUEUER = CompletedAssignment("enqueuer")
 FORMATTING = CompletedAssignment("formatting", (ENQUEUER,))
-FIRST_EDITOR = CompletedAssignment("first_editor", (FORMATTING,))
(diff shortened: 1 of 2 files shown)


rudimatz:
also address issue as discussed with @ajeanmahoney that a pending activity is displayed while ref_checker is not done yet.
`ref_checker` may happen in parallel with formatting, but `first_edit` shall only happen once both formatting and ref_checker are done.

image

see https://github.com/ietf-tools/queue/issues/109#issuecomment-5281978438

jennifer-richards (approved):

PR #1460, opened 2026-08-17
feat: update subject line and add cc

### LINKED ISSUE #1458: for Send Email, Announce Publication: add subseries number to subject line
For a given document, if there's an BCP or STD number (set by the user in the Info tab in "Subseries" field), when the user creates a publication announcement (i.e., clicks Send Email, then Announce Publication), please include the subseries number (and comma) at the start of the subject line. 

Subject line template:
* current: RFC XXXX on [RFC title]
* desired: BCP NNN, RFC XXXX on [RFC title]  - or - 
STD NNN, RFC NNNN on [RFC title]

Examples:
* current: RFC 10027 on Best Current Practice for Security of Cross-Device Flows
* desired: BCP 247, RFC 10027 on Best Current Practice for Security of Cross-Device Flows 
* actual pub announcements: https://mailarchive.ietf.org/arch/msg/rfc-dist/1ON-5Hd-o0Vkjoe7ClNLj_hPJnc/ and https://mailarchive.ietf.org/arch/msg/rfc-dist/Q64ceosKkNJjUMjYMJj7uG9cy6I/

[current workaround: update manually]

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1458 and add iesg_contact and stream_manager to CCs for ALL email types for IETF non-wg docs (as requested by @sginoza)

### DIFF
--- a/rpc/api.py
+++ b/rpc/api.py
@@ -2667,6 +2667,12 @@ class RfcMailTemplatesList(views.APIView):
             if shepherd_email:
                 interested_parties.add(shepherd_email)
         stream_slug = rfc_to_be.stream_id
+
+        individual_ietf_contacts: set[str] = set()
+        if stream_slug == "ietf" and not rfc_to_be.group:
+            for person in (rfc_to_be.iesg_contact, rfc_to_be.stream_manager):
+                if person is not None and person.email:
+                    individual_ietf_contacts.add(person.email)
         if stream_slug == "iab":
             interested_parties.add("iab@iab.org")
         elif stream_slug == "ise":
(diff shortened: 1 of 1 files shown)


rpc/api.py
```diff
@@ -2696,12 +2704,19 @@ def get(self, request, rfctobe_id, format=None):
                 ietf_list_email = datatracker_group_list_email(rfc_to_be.group)
                 if ietf_list_email:
                     publication_cc.add(ietf_list_email)
+            else:
+                publication_cc |= individual_ietf_contacts
```
jennifer-richards: Just an observation: there is a long string of `if` statements branching on `stream_slug` - might be worth a refactor to consolidate those. It takes a lot of staring to trace the logic to see what the overall treatment is for a particular stream.

jennifer-richards (approved):
One comment inline, but it's not a request for this PR

rpc/api.py
rudimatz: indeed, this grew too wild. will simplify the logic!

PR #1466, opened 2026-08-18
feat: allow full github url for repo

### LINKED ISSUE #1438: Changing the repository field to hold the entire URL
Per the team, it is easier to copy and paste the full URL than it is to hold a portion of it.  Can we update this field to hold the full URL?

Old: rfc-editor-drafts/draft-ietf-tls-ecdhe-mlkem

New: https://github.com/rfc-editor-drafts/draft-ietf-tls-ecdhe-mlkem

rudimatz:
resolve https://github.com/ietf-tools/purple/issues/1438

### DIFF
--- a/rpc/lifecycle/repo.py
+++ b/rpc/lifecycle/repo.py
@@ -3,6 +3,7 @@
 
 import json
 import logging
+import re
 from pathlib import PurePath
 
 import jsonschema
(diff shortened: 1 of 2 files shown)


rudimatz:
logic: only store the relative path in DB as happening currently. Normalize URL before writing. Show error when string is invalid

jennifer-richards (approved):

PR #1467, opened 2026-08-18
feat: improve dark mode colors visibility

### LINKED ISSUE #1282: Dark mode is too dark when entering text in "Comments" box
I can barely see the text I'm typing in the comment box when using Purple's dark mode:

Image

Tested with system settings set to "Light", "Dark", and "Auto" on macOS (Sequoia 15.7.4) but no change with the issue.

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1282

### DIFF
--- a/client/app/components/DefaultLayoutAfterAuthWall.vue
+++ b/client/app/components/DefaultLayoutAfterAuthWall.vue
@@ -85,6 +85,23 @@ body {
 .dark body {
   background-color: #0a0a0a;
 }
+/* Dark-mode fallback for form controls that don't set their own colors.
+   :where() keeps specificity minimal so components with explicit dark: classes win. */
+.dark textarea,
+.dark select,
+.dark input:where(:not([type='checkbox']):not([type='radio'])) {
+  color: #f3f4f6;
+}
+/* Field background only for standalone controls; textareas keep a transparent
+   background so they blend with their own container (e.g. the comment box). */
+.dark select,
+.dark input:where(:not([type='checkbox']):not([type='radio'])) {
+  background-color: #262626;
+}
+.dark input::placeholder,
+.dark textarea::placeholder {
+  color: #9ca3af;
+}
 #__nuxt {
   min-height: 100vh;
 }


client/app/components/DefaultLayoutAfterAuthWall.vue
```diff
@@ -85,6 +85,23 @@ body {
...
+   :where() keeps specificity minimal so components with explicit dark: classes win. */
+.dark textarea,
+.dark select,
+.dark input:where(:not([type='checkbox']):not([type='radio'])) {
+  color: #f3f4f6;
+}
+/* Field background only for standalone controls; textareas keep a transparent
+   background so they blend with their own container (e.g. the comment box). */
+.dark select,
+.dark input:where(:not([type='checkbox']):not([type='radio'])) {
```
holloway, line 98: should this selector also include textarea?
```suggestion
..dark textarea,
```

client/app/components/DefaultLayoutAfterAuthWall.vue
```diff
@@ -85,6 +85,23 @@ body {
 .dark body {
   background-color: #0a0a0a;
 }
+/* Dark-mode fallback for form controls that don't set their own colors.
+   :where() keeps specificity minimal so components with explicit dark: classes win. */
+.dark textarea,
```
holloway, line 90: Have you tested whether you can override these with tailwind css?

I suspect you might need `:where(.dark) textarea` (etc) to reduce specificity for that to work.

client/app/components/DefaultLayoutAfterAuthWall.vue
rudimatz: see comment above... it looks better when being transparent

client/app/components/DefaultLayoutAfterAuthWall.vue
rudimatz: Confirmed the dark-mode input fallback is overridable: a component's dark: utility wins over it

holloway (approved):
Seems like a good default, to override later with tailwind classes

PR #1469, opened 2026-08-19
feat: show expedited docs always on top

### LINKED ISSUE #1435: Prioritize queue display for expedited docs
In the queue <https://purple.rfc-editor.org/queue/queue>, please pin expedited docs to the top of the list so we can easily keep track of the expedited documents (docs with "expedited" label).   We published RFC 10026, which had been expedited.  It was easy to lose track of the expedited document in the queue.  Having it appear at the top of the queue would ensure we're seeing the current status each time we look at the queue page.

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1435

### DIFF
--- a/client/app/pages/queue/queue.vue
+++ b/client/app/pages/queue/queue.vue
@@ -126,7 +126,10 @@
             :error="[error, clustersError, peopleError]"
             :column-count="table.getAllColumns().length"
             :row-count="table.getRowModel().rows.length" />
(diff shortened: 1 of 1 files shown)


rudimatz:
as per linked discussion, agreed-upon solution:
set default order logic:: sort by enqueue date ASC

display for docs in queue:

1) show expedited on top
2) sort the rest by oldest in queue first

jennifer-richards (approved):

PR #1474, opened 2026-08-20
feat: use role names instead of slugs

rudimatz:
relates to the work in https://github.com/ietf-tools/purple/pull/1421

for assignment roles, display names, not slugs

### DIFF
--- a/client/app/components/DocumentCard.vue
+++ b/client/app/components/DocumentCard.vue
@@ -18,7 +18,7 @@ Based on https://tailwindui.com/components/application-ui/lists/grid-lists#compo
         </div>
       </Anchor>
       <div v-for="role in cookedDocument.needsAssignment">
-        <BaseBadge :label="`Needs ${role?.name ?? '(unnamed)'}`" />
+        <BaseBadge :label="`Needs ${role.name}`" />
       </div>
       <HeadlessMenu as="div" class="relative ml-auto">
         <HeadlessMenuButton class="-m-2.5 block p-2.5 text-gray-400 hover:text-gray-500">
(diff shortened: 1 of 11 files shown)


client/app/components/DocumentCard.vue
```diff
@@ -18,7 +18,7 @@ Based on https://tailwindui.com/components/application-ui/lists/grid-lists#compo
         </div>
       </Anchor>
       <div v-for="role in cookedDocument.needsAssignment">
-        <BaseBadge :label="`Needs ${role ?? '(unnamed)'}`" />
+        <BaseBadge :label="`Needs ${role.name}`" />
```
jennifer-richards, line 21: Is the guard against a missing role not still needed?

jennifer-richards (approved):
Seems ok. It makes apparent that roles are sometimes passed as slugs and sometimes passed as objects. If internal purple APIs aren't all consistent on using the slug and counting on the `useRoleName` composable, then it's worth considering future work to unify the approach.

client/app/components/DocumentCard.vue
rudimatz: no, I validated. it's in a for-loop for each role, and role.name is required field (not blank)... so I removed the guards

rudimatz:
> Seems ok. It makes apparent that roles are sometimes passed as slugs and sometimes passed as objects. If internal purple APIs aren't all consistent on using the slug and counting on the useRoleName composable, then it's worth considering future work to unify the approach.

yes maybe... here I made the choice that within RfcToBe we use objects consistently ... but other instances might use slug+composable. It's a trade-off... but it also reduces payloads. Since we were needing the composable at some places anyways, I thought its better to use it throughout, rather than changing the serializer.

But we can always review

jennifer-richards (approved):

PR #1475, opened 2026-08-20
feat: add notification system (for blocks)

### LINKED ISSUE #1426: Notify when a document moves to In Progress because their norm refs are in the queue
When a document becomes unblocked because all of its normative references are now in the queue, purple should send email to rfc-editor@rfc-editor.org saying that document has moved from Blocked Reference Not Received to In Progress, and noting what the pending activity is:
```
To: rfc-editor@rfc-editor.org
Subject: <draft-ietf-teep-protocol> now In Progress

Normative reference <draft-ietf-suit-manifest> has entered the queue. 
<draft-ietf-teep-protocol> is now In Progress. 

Awaiting formatting.
```

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1426

implement a simple notification/broadcast system:

- new Notifications are currently created when a draft becomes unblocked
- System allows for broadcasts (send to all) or individual notifications per user
- a bell on top right of screens indicates if there are `new` notifications
- a `notification` page lists all past notifications

System is versatile and can be user for more use-cases later

### DIFF
--- a/client/app/composables/useNotifications.ts
+++ b/client/app/composables/useNotifications.ts
@@ -26,7 +26,7 @@ export const useNotifications = () => {
 
   onBeforeMount(() => {
     instanceCount++
-    if (userStore.authenticated) poll()
+    if (userStore.authenticated) void poll()
     if (interval === null) {
       interval = setInterval(poll, POLL_MS)
     }
--- a/client/app/composables/useNotifications.ts
+++ b/client/app/composables/useNotifications.ts
@@ -1,0 +1,53 @@
+import { ref, onBeforeMount, onUnmounted, readonly } from 'vue'
+import type { PurpleApi } from '~/purple_client'
+import { useUserStore } from '~/stores/user'
+
+// Shared singleton: one poll drives the header bell no matter how many
+// components read the count (mirrors the useCurrentTime pattern).
+const POLL_MS = 60_000
+const unreadCount = ref(0)
+let interval: ReturnType<typeof setInterval> | null = null
+let instanceCount = 0
+let api: PurpleApi | null = null
+
+const poll = async () => {
+  if (!api) return
+  try {
+    const { count } = await api.notificationsUnreadCount()
+    unreadCount.value = count
+  } catch {
+    // transient network/auth errors: keep the last known count
+  }
+}
+
+export const useNotifications = () => {
+  api = useApi()
+  const userStore = useUserStore()
+
+  onBeforeMount(() => {
+    instanceCount++
+    if (userStore.authenticated) poll()
+    if (interval === null) {
+      interval = setInterval(poll, POLL_MS)
+    }
+  })
+  onUnmounted(() => {
+    instanceCount--
+    if (instanceCount <= 0 && interval !== null) {
+      clearInterval(interval)
+      interval = null
+    }
+  })
+
+  const markAllRead = async () => {
+    if (!api) return
+    await api.notificationsMarkRead()
+    unreadCount.value = 0
+  }
+
+  return {
+    unreadCount: readonly(unreadCount),
+    refresh: poll,
+    markAllRead
+  }
+}
--- a/client/app/composables/useNotifications.ts
+++ b/client/app/composables/useNotifications.ts
@@ -1,3 +1,5 @@
+// Copyright The IETF Trust 2026, All Rights Reserved
+
 import { ref, onBeforeMount, onUnmounted, readonly } from 'vue'
 import type { PurpleApi } from '~/purple_client'
 import { useUserStore } from '~/stores/user'
@@ -41,8 +43,12 @@ export const useNotifications = () => {
 
   const markAllRead = async () => {
     if (!api) return
-    await api.notificationsMarkRead()
-    unreadCount.value = 0
+    try {
+      await api.notificationsMarkRead()
+      unreadCount.value = 0
+    } catch {
+      // leave the count as-is; the next poll re-confirms the true value
+    }
   }
 
   return {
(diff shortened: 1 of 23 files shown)


rudimatz:
displays a bell that will link to a new notification page
image

rpc/api.py
```diff
@@ -1273,6 +1276,80 @@ class AssignmentViewSet(viewsets.ModelViewSet):
...
+            "NotificationUnreadCount", fields={"count": serializers.IntegerField()}
+        ),
+    )
+    @action(detail=False, methods=["get"], url_path="unread_count")
+    def unread_count(self, request):
+        person = self._rpcperson()
+        if person is None:
+            return Response({"count": 0})
+        seen_at = self._seen_at(person)
+        unread = Q() if seen_at is None else Q(created__gt=seen_at)
```
jennifer-richards: Just a style comment (not even a nit really): IMO this would be easier to read if it were written out as something like `unread = self.get_queryset() if seen_at is None else self.get_queryset().filter(created_at...)` instead of by passing `Q()` objects around. (Really, I'd probably write it out as a multi-line if statement)

I'm not really asking for a change, just noticed this and have been bitten by operations on Q objects being counterintuitive at times. For readability / maintainability, sometimes it's better to be a little wordier or use additional lines of code.

rpc/models.py
```diff
@@ -1470,3 +1470,81 @@ class Slugs(models.TextChoices):
...
+    rfc_to_be = models.ForeignKey(
+        "RfcToBe",
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE,
+        help_text="Document this notification is about",
+    )
+    # Denormalized event detail (draft name, blocking reason names) so a
+    # notification renders without re-deriving state that may since have changed.
+    data = models.JSONField(default=dict, blank=True)
```
jennifer-richards: Might consider separately modeling `draft_name` (as a char field) and `reasons` (as an arrayfield or a JSONField that's a list) to reduce the amount of denormalized schema. The serializer code is depending on the structure of this denormalized data which is a hint that a JSONField is creating a hidden maintenance burden.

client/app/composables/useNotifications.ts
```diff
@@ -0,0 +1,53 @@
+import { ref, onBeforeMount, onUnmounted, readonly } from 'vue'
```
jennifer-richards, line 3: missing copyright stmt

client/app/composables/useNotifications.ts
```diff
@@ -0,0 +1,53 @@
...
+    instanceCount--
+    if (instanceCount <= 0 && interval !== null) {
+      clearInterval(interval)
+      interval = null
+    }
+  })
+
+  const markAllRead = async () => {
+    if (!api) return
+    await api.notificationsMarkRead()
```
jennifer-richards: unhandled exception opportunity - is there a sensible thing to do if this API call fails?

rpc/api.py
```diff
@@ -1273,6 +1276,80 @@ class AssignmentViewSet(viewsets.ModelViewSet):
...
+        return self._rpcperson_cache
+
+    def _seen_at(self, person):
+        marker = NotificationReadMarker.objects.filter(person=person).first()
+        return marker.seen_at if marker else None
+
+    def get_queryset(self):
+        user = self.request.user
+        if not user.is_authenticated:
+            return Notification.objects.none()
```
jennifer-richards, line 1309: I don't think this API is exposed for non-authenticated users. Might consider raising an exception or emitting a log if a non-authenticated user gets here as an early warning that something is not working as expected. (OTOH that might be better covered by a test that does a more systematic confirmation that our auth system is operating as intended and doing it at an API-by-API level is not desirable)

jennifer-richards (commented):
A few nits / comments inline.

rpc/models.py
rudimatz: I refactored the schema, will now use a "message" field that is rendered on creation and later needs just to be displayed. No more need for denormalized data in JSONField

rpc/api.py
rudimatz: ok I'll add test that will test the auth system globally

rudimatz:
force-push was unavoidable... too many changes on main

rpc/tests.py
```diff
@@ -739,6 +739,47 @@ def test_only_unblock_emits_broadcast_notification(self):
...
+            ("get", "/api/rpc/notifications/"),
+            ("get", "/api/rpc/notifications/unread_count/"),
+            ("post", "/api/rpc/notifications/mark_read/"),
+        ):
+            resp = getattr(self.client, method)(path)
+            self.assertIn(resp.status_code, (401, 403), (path, resp.status_code))
+
+
+class AuthEnforcementTests(TestCase):
+    """Guard against a regression that opens the API to anonymous users."""
```
jennifer-richards: As I understand it, these tests are proving that the permissions we're using for viewsets are what we intended, relying on the `-list` endpoint existing in all cases and being safe / easy to check. That's great, no complaint, just checking that we agree.

It'd be good to capture this in the docstring. This test isn't proving that every endpoint that exists is guarded. Ones done other than via viewsets, or any that might someday be added with custom permissions checks would need their own test cases added alongside the ones that are created here. That'll help human understanding and act as a breadcrumb to help future AI work organize tests appropriately.

jennifer-richards (changes requested):
One change suggested - just a docstring adjustment, but I think it's important

rpc/tests.py
rudimatz: OK I'll rename test and update comment

jennifer-richards (approved):

PR #1482, opened 2026-08-25
feat: improve behavior blocking refs

### LINKED ISSUE #1477: Cluster doc should be blocked but isn't
An editor completed the first edit of a C517 document with a normative reference (draft-ietf-teas-ietf-network-slice-nbi-yang). This document should show blocked with "Reference First Edit Incomplete" but it doesn't.

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1477

Problem:
A normative ("refqueue") reference created before its target entered the queue points at the datatracker Document, with target_rfctobe null. The second-edit and publish gates dereferenced target_rfctobe unconditionally, so such a reference raised AttributeError and aborted block evaluation — leaving the citing document silently unblocked. The import flow also flipped the relation type not-received → refqueue without re-linking the target, which is what produced those rows in the first place.

Changes:
- re-link reference from DT-document to the newly created RfcToBe at import
- a refqueue reference that doesn't resolve to a queued RfcToBe now blocks the citing doc (REFQUEUE_FIRST_EDIT_INCOMPLETE / REFQUEUE_PUBLISH_INCOMPLETE) instead of crashing — "not in the queue" is treated as "not ready".
- Backfill (0015_backfill_refqueue_target_rfctobe): re-links existing stale rows, and writes simple-history snapshots for each change so it isn't later misattributed to the next editor (migrations bypass simple-history otherwise).
- add tests

### DIFF
--- a/rpc/lifecycle/blocked_assignments.py
+++ b/rpc/lifecycle/blocked_assignments.py
@@ -97,6 +97,7 @@ def get_block_reasons(rfc: RfcToBe) -> set[str]:
             for ref in refqueue_qs:
                 target = ref.target_rfctobe
                 if (
+                    # Not in the queue, so not published.
                     target is None
                     or target.incomplete_activities()
                     .filter(slug="first_editor")
--- a/rpc/lifecycle/blocked_assignments.py
+++ b/rpc/lifecycle/blocked_assignments.py
@@ -95,8 +95,10 @@ def get_block_reasons(rfc: RfcToBe) -> set[str]:
         refqueue_qs = rfc.rpcrelateddocument_set.filter(relationship="refqueue")
         if refqueue_qs.exists():
             for ref in refqueue_qs:
+                target = ref.target_rfctobe
                 if (
-                    ref.target_rfctobe.incomplete_activities()
+                    target is None
+                    or target.incomplete_activities()
                     .filter(slug="first_editor")
                     .exists()
                 ):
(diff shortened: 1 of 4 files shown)


rpc/lifecycle/blocked_assignments.py
```diff
@@ -95,8 +95,10 @@ def get_block_reasons(rfc: RfcToBe) -> set[str]:
         refqueue_qs = rfc.rpcrelateddocument_set.filter(relationship="refqueue")
         if refqueue_qs.exists():
             for ref in refqueue_qs:
+                target = ref.target_rfctobe
```
jennifer-richards, line 98: nit: suggest a brief comment explaining the logic here (i.e., that lack of a `target_rfctobe` -> not in the queue yet) as was done below

rpc/migrations/0015_backfill_refqueue_target_rfctobe.py
```diff
@@ -0,0 +1,82 @@
...
+
+    A refqueue row created before its target entered the queue points at the
+    datatracker Document. Once the target has an RfcToBe, the blocking gates and
+    the signal-driven re-evaluation both key off target_rfctobe, so such rows are
+    invisible to them. Convert each to point at the active RfcToBe (exactly one
+    target may be set, so target_document is cleared). Drop rows that would
+    duplicate an existing target_rfctobe reference.
+    """
+    RpcRelatedDocument = apps.get_model("rpc", "RpcRelatedDocument")
+    Historical = apps.get_model("rpc", "HistoricalRpcRelatedDocument")
```
jennifer-richards: nit: `HistoricalRpcRelatedDocument` for naming consistency

jennifer-richards (approved):
a couple nits, up to you whether to accept them or not

PR #1485, opened 2026-08-25
feat: add filter reset button on queue page

### LINKED ISSUE #1480: Queue page: add ability to clear filters with a click
When a user uses filters to narrow down the queue list, the have to click each filter to reset them. There should be a way to clear all set filters.

rudimatz:
resolve https://github.com/ietf-tools/purple/issues/1480


rudimatz:
image

jennifer-richards (approved):

PR #1487, opened 2026-08-26
feat: improve exception logging

### LINKED ISSUE #1471: Exception logging may omit information about the exception
In some places (e.g., https://github.com/ietf-tools/purple/blob/6bbd28196aa76f7845c7b5265d36d8f1bccdcbe5/rpc/tasks.py#L227), exceptions are logged with a statement like
```python
    try:
        notify_datatracker_queue()
    except Exception as e:
        logger.error(f"Error in push_queue_to_datatracker_task: {e}")
```
This works when the exception renders an informative string, but this does not work for many situations. E.g., for a bare `ValueError`
```python
>>> f"Error: {ValueError()}"
'Error: '
```

These should probably turn into something that calls `repr()`. That could be `f"{e!r}"` or `f"{repr(e)}"`, which are equivalent and both produce
```python
>>> f"Error: {ValueError()!r}"
'Error: ValueError()'
```

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1471

### DIFF
--- a/rpc/tasks.py
+++ b/rpc/tasks.py
@@ -173,8 +173,8 @@ def validate_metadata_task(self, rfc_to_be_id):
         _save_metadata_results(rfc_to_be, head_sha, metadata, status)
 
     except Exception as e:
-        logger.error(f"Error in validate_metadata_task: {e}")
-        detail = str(e)
+        logger.exception("Error in validate_metadata_task")
+        detail = str(e) or repr(e)
         status = MetadataValidationResults.Status.FAILED
         _save_metadata_results(rfc_to_be, head_sha, metadata, status, detail)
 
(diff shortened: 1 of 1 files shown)


rpc/tasks.py
```diff
@@ -173,8 +173,8 @@ def _save_metadata_results(rfc_to_be, head_sha, metadata, status, detail=None):
         _save_metadata_results(rfc_to_be, head_sha, metadata, status)
 
     except Exception as e:
-        logger.error(f"Error in validate_metadata_task: {e}")
-        detail = str(e)
+        logger.exception("Error in validate_metadata_task")
+        detail = str(e) or repr(e)
```
rudimatz, line 177: @jennifer-richards note this code follows same logic as existing https://github.com/ietf-tools/purple/pull/1487/changes#diff-ca509616531c8f853d0b74fed4df7e2e78d5fa5ec4140f292e8f64cae2e6eee7L192

jennifer-richards (approved):

PR #1489, opened 2026-08-26
feat: add fields for label: name, desc

### LINKED ISSUE #1344: Edit Label: Provide a field for capturing details
The Edit Label modal should have a "Details" field where the user can provide more information about the label. These details can then be displayed when a user mouses over the label when it is shown on other parts of purple and queue.rfc-editor.org. See https://github.com/ietf-tools/queue/issues/101

### LINKED ISSUE #1486: for labels add NAME and DESCRIPTION
labels should have, besides `slug`, a NAME and DESCRIPTION

slug should be auto-generated from name, but not editable by users.
Name and Desc might be updated later.

References in code, e.g. for blocks, should refer to the slug.

relates to https://github.com/ietf-tools/purple/issues/1344

this will require migration to fill all existing labels

rudimatz:
resolve https://github.com/ietf-tools/purple/issues/1486 and resolve https://github.com/ietf-tools/purple/issues/1344

Labels should have a slug (unique internal identifier) and a `name` that is displayed user facing.

**Once a new label is created, the `slug` and `name` can't be changed via UI.**

A migration file will add the new fields and create slugs for existing labels.
Logic in code will refer to slugs (snake_case), while label names can be adjusted later via UI


rudimatz:
this will require an compagnon PR in queue to display label.name instead of label.slug
related to https://github.com/ietf-tools/queue/pull/133

Order:

- deploy purple
- build and deploy updated precomputer
- run Tekton
- update queue website

rudimatz:
added text to explain label name changes
image

rpc/migrations/0016_label_name_description.py
```diff
@@ -0,0 +1,108 @@
...
+    "is_exception",
+    "is_complexity",
+    "color",
+    "used",
+    "is_public",
+)
+
+
+def backfill_label_name_and_slug(apps, schema_editor):
+    """Move each label's old slug into name and re-slug it to kebab-case.
```
jennifer-richards: That's a lie! The code goes out of its way to avoid kebab-case, preferring underscores.

(Personally I prefer kebab to snake for slugs, so I'd be happy if the fix was to keep the comment and change everything else, but either way is ok)

rpc/models.py
```diff
@@ -1208,7 +1208,13 @@ class Label(models.Model):
     ### Will have to have LabelHistory on objects that have collections of labels
     ### That is, we need to compute when something had a label and how long
 
-    slug = models.CharField(max_length=64, unique=True)
+    slug = models.CharField(
+        max_length=64,
+        unique=True,
+        help_text="Stable machine key, auto-generated from name; referenced in code.",
+    )
+    name = models.CharField(max_length=255, default="", help_text="Human-readable name")
```
jennifer-richards: Bit of a naming nit - calling this the "name" of the label seems confusing to me. It's the label text more than the label name. The slug is closer semantically to the name of the label.

Not a demand for change, just an observation.

jennifer-richards (changes requested):
Couple comments inline - one is an inaccurate comment (or inaccurate code, depending on how you resolve it) and the other is a bigger naming change to consider but I don't insist.

Also - it'd be nice if the migration verified that it is going to produce slugs corresponding to the values embedded in the code. I.e., if somehow there got to be a mismatch between the labels that exist and what it expects, this might leave things broken. It'd be acceptable to verify this after the upgrade, but imo it'd be better to fail hard if an assumption like that isn't accurate.

rudimatz:
@jennifer-richards I renamed `label.name` to `label.text`, and made the slug for labels kebab case

jennifer-richards (approved):

PR #1502, opened 2026-09-03
feat: tweak schema and show dash if date is null

### LINKED ISSUE #1501: fix null date displays
Image

if a date is null, show more graceful data in UI

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1501

enqueued_at should be not null and I validated that it is always set for all docs


jennifer-richards (approved):
Looks good and fixes a display wart.

I think we should probably more aggressively flag the absence of a submitted date. It shouldn't ever be null (and only came up because of [a datatracker bug](https://github.com/ietf-tools/datatracker/issues/11701)).

PR #1503, opened 2026-09-04
feat: add rev check to metadata compare

### LINKED ISSUE #1499: include "rev" in metadata compare before publish
following discussion with @ajeanmahoney and @jennifer-richards , adding this as new feature request:

`rev` field shall be included in the metadata compare step before publication, since a mismatch will later cause error

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1499

### DIFF
--- a/rpc/lifecycle/metadata.py
+++ b/rpc/lifecycle/metadata.py
@@ -5,11 +5,13 @@ import datetime
 import logging
 import re
 import xml.etree.ElementTree as ET
+from functools import cached_property
 from itertools import zip_longest
 from typing import Any
 
 from django.db import transaction
 
+from datatracker.rpcapi import with_rpcapi
 from rpc.models import (
     DocRelationshipName,
     RfcToBe,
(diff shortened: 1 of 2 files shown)


rudimatz:
logic used for 3-way compare of `revision`:

DB == RFCXML == DT => ALL FINE
DB !== RFCXML == DT => NO match but offer fix (update DB with value from DT/XML)
RFCXML !== DT => hard error, not fixable, needs manual intervention

jennifer-richards (changes requested):
I think this should deal with `RfcToBe`s that don't have drafts - we're going to need to support those soon. There's no datatracker doc or revision to check in that case.

The UI should distinguish "datatracker doesn't know the doc name" from "datatracker threw an error on the API call." Right now, if I trace the `None` handling correctly, they both just show up as "couldn't find the rev."

Dropping down to the nit level, the `agreed_rev` property has a strange name to me. I was about to suggest inlining the code in `compare_revision()` because having it factored out makes it hard to see the logic involved, but then I saw it's used earlier in the diff. I don't have a great suggestion for a better name, but maybe there's something out there.

rudimatz:
@jennifer-richards look at that nice Merge commit, avoiding force push.
in latest commit: I added output to show more detailed message if DT returned error. For draft-less RFCs, we only compare the RFCXML with current DB value, skipping the 3 way compare

jennifer-richards (approved):

Issue #1468, opened 2026-08-18
Assignment info should be visible on xfer docs


ajeanmahoney:
Documents that were published before cutover to Purple are missing assignment information. For example, https://purple.rfc-editor.org/docs/draft-ietf-dmarc-failure-reporting/assignments shows no assignment information. Its assignment data from rfced db is "SS/AP/KC/SG" so it should show at least the following:

Sarah Tarrant    formatting done
Alanna Paloma    first_editor done
Karen Moore    second_editor done
Alanna Paloma    final_review_editor done
Sandy Ginoza    publisher done

rudimatz:
@ajeanmahoney following this logic, is the first editor always same as final review editor?

ajeanmahoney:
Not always. When editors are being trained, they don't handle final reviews -- the 2nd editor handles them. But you can't see that from the old database assignments (there may be comments saying the RE handled AUTH48 but that may not have been consistently noted).
So maybe we shouldn't note the final review editor because we can't know for sure.

rudimatz:
how about "enqueuer" or "ref_checker", should that also be set, as we did for the in-queue docs?

ajeanmahoney:
The old system did not have a way to store "ref_checker" information once the document moved to RE (2nd editor), so docs published before the cut over to Purple don't have that info. For the most part, the person noted in the old db's CE field did enqueuing and formatting. The following initials indicate a separation of those tasks:
CT - Chris did pre-EDIT, Tess did formatting 
ST - Sarah did pre-EDIT, Tess did formatting

rudimatz:
since 9 former RPC persons don't have DT accounts, we start to tackle towards backfilling for existing (active or inactive) RPC persons in purple. (We can Add Tess who has an account).

This will cover the biggest chunk of existing assignment data

ajeanmahoney:
yes, this should focus on the current RPC team members' past assignments.

rudimatz:
[rfc_assignments_backfill.csv](https://github.com/user-attachments/files/31978903/rfc_assignments_backfill.csv)

legacy assignments to be backfilled (spot check in progress by @ajeanmahoney )

ajeanmahoney:
Let's just backfill assignments that were made from 1 January 2026 onward.

rudimatz:
adding comment:
[...]

PR #1512, opened 2026-09-08
fix: never block published docs

rudimatz:
relates to https://github.com/ietf-tools/purple/issues/1468


rudimatz:
background: the related task will backfill 15.000 assignments. we want to avoid those possibly causing blocks, where it doesn't make sense

Adding the guard makes the existing logic more consistent

jennifer-richards (approved):

PR #1514, opened 2026-09-08
feat: backfill legacy assignments

### LINKED ISSUE #1468: Assignment info should be visible on xfer docs

rudimatz:
fix https://github.com/ietf-tools/purple/issues/1468


rudimatz:
expected output (dry run)

```
INFO 2026-09-08 21:45:09,432 Task rpc.backfill_rfc_assignments[6ad4a699-e78e-4e2e-85bb-d85ba519dfb5] succeeded in 0.049134211003547534s: {'rows': 15881, 'already_present': 122, 'to_create': 15759, 'unresolved': 0, 'unresolved_examples': [], 'created': 0, 'dry_run': True}
```

set `{"dry_run": false}` on periodic task to RUN for real

rudimatz:
once done, this code may be reverted/removed

rpc/tasks_backfill_assignments.py
```diff
@@ -0,0 +1,1544 @@
...
+from celery import shared_task
+from celery.utils.log import get_task_logger
+from django.db import transaction
+
+from .models import Assignment, RfcToBe, RpcPerson, RpcRole
+from .signals import SignalsManager
+
+logger = get_task_logger(__name__)
+
+CHANGE_REASON = "backfill: assignments transferred from the rfced database (#1468)"
```
jennifer-richards, line 42: I don't think adding the issue number to the change reason is necessary (but if you want it, it should explain what the number means for the benefit of someone in a future where we're using a different system for issue tracking)

rpc/tasks_backfill_assignments.py
```diff
@@ -0,0 +1,1544 @@
...
+            9171, 9176, 9189, 9198, 9204, 9231, 9299, 9304, 9305, 9344, 9351, 9361,
+            9365, 9367, 9375, 9376, 9390, 9392, 9424, 9439, 9462, 9486, 9528, 9582,
+            9584, 9594, 9601, 9621, 9622, 9623, 9672, 9681, 9689, 9692, 9696, 9703,
+            9746, 9849, 9948, 9949],
+    },
+}
+# fmt: on
+
+
+@shared_task(name="rpc.backfill_rfc_assignments", bind=True)
```
jennifer-richards, line 1458: Is the explicit naming needed because of the funny way this is imported? Mostly I'm just curious, but keep in mind that explicit naming makes it more likely we end up with a collision vs letting things pick up names based on the actual code structure.

rpc/tasks_backfill_assignments.py
```diff
@@ -0,0 +1,1544 @@
...
+    what each run did, alongside the per-row history the writes leave behind.
+    """
+    wanted_rows = [
+        (rfc, dt, role)
+        for (dt, _name), roles in ASSIGNMENTS.items()
+        for role, rfcs in roles.items()
+        for rfc in rfcs
+    ]
+    if limit:
+        wanted_rows = wanted_rows[: int(limit)]
```
jennifer-richards, line 1472: Since the `ASSIGNMENTS` structure is fixed, this will always choose the same limited number of rows rather than allow stepping through a bit at a time. Is that intentional?

rpc/tasks_backfill_assignments.py
```diff
@@ -0,0 +1,1544 @@
...
+            with transaction.atomic():
+                for rfc_to_be_id, pid, role in wanted[start : start + CHUNK]:
+                    assignment = Assignment(
+                        rfc_to_be_id=rfc_to_be_id,
+                        person_id=pid,
+                        role_id=role,
+                        state=Assignment.State.DONE,
+                    )
+                    assignment._change_reason = CHANGE_REASON
+                    assignment.save()
```
jennifer-richards, line 1538: Just to confirm, `save()` with `SimpleHistory` still works with signals off?

jennifer-richards (changes requested):
One or two nits (depending on your preference) plus a couple questions.

rpc/tasks_backfill_assignments.py
rudimatz: did cleanup, the number is indeed not useful

rpc/tasks_backfill_assignments.py
rudimatz: yes

rpc/tasks_backfill_assignments.py
rudimatz: the cap now applies to the rows still missing, so repeated limited runs step through the set

rudimatz:
changed logic to only import assignments that were created in legacy DB before Jan 1st 2026. (as per @ajeanmahoney 's request)

Script will import 496 new assignments (from beginning of 2026 until cutover)

jennifer-richards (approved):

PR #1533, opened 2026-09-15
feat: add editorial notes

### LINKED ISSUE #1491: Add an "Editorial Notes" tab for document pages
Please add a tab to a document's page called "Editorial Notes". This tab will contain a large text field that can be used as a scratch pad for the editors. 

Image

An editor clicks the Edit button to make updates, and clicks Save when done. Another editor can add, change, delete contents when they click the Edit button.

(Note: The "Search messages" button shown in the wireframe will be covered in another feature request.)

Not required:

- Keeping history. That is, this doesn't need to be a wiki page, and it doesn't need an edit log.
- Markdown support. Although the team would like URLs to be clickable when viewing contents. The team doesn't want to learn another flavor of markdown. So if it's easy to support GFM markdown, then that flavor is okay, or if it's easy to add formatting buttons, that would also work.

Not required, yet:

- After experimenting with this, the team may want the field to be pre-populated with a template (adding headings such as Actions, Questions, etc.). Just something to keep in mind when implementing.

### LINKED ISSUE #1492: Add the ability to browse the rfc-editor mailarchive for a particular doc
This is very similar to the "Search email archive" button on an I-D's dt status page. However, the search should be narrowed to the following:

- rfc-editor@rfc-editor.org mailing list
- Subject line contains the document's draftstring

This "Search messages" button should reside on the document's "Editorial Notes" tab (see https://github.com/ietf-tools/purple/issues/1491). When the user clicks the button, the mailarchive page should be launched in a separate browser tab.

rudimatz:
resolve https://github.com/ietf-tools/purple/issues/1491 and resolve https://github.com/ietf-tools/purple/issues/1492


rudimatz:
preview 
image

client/app/pages/docs/[id]/editorial-notes.vue
```diff
@@ -0,0 +1,138 @@
...
+          <template v-if="note.updatedBy">by {{ note.updatedBy.name }}</template>
+        </p>
+
+        <textarea
+          v-if="isEditing"
+          v-model="draftText"
+          rows="20"
+          aria-label="Editorial notes"
+          class="mt-4 block w-full rounded border border-gray-300 bg-white px-3 py-2 text-sm text-black focus:outline-none focus:ring-2 focus:ring-blue-500 dark:bg-black dark:text-white"
+          placeholder="Notes for the editors working on this document" />
```
Comment (jennifer-richards) [inline client/app/pages/docs/[id]/editorial-notes.vue]: ux nit: not sure the placeholder is appropriate. User is probably familiar with a text box and they just opened it with the 'edit' button. Seems like noise to me.

If the note about the audience for the note is needed, it should be shown somewhere that is visible even when there's text in the field.

client/app/utils/linkify.ts
```diff
@@ -0,0 +1,37 @@
...
+    trimmed = trimmed.slice(0, -1).replace(TRAILING_PUNCTUATION, '')
+  }
+  return trimmed
+}
+
+/**
+ * Split plain text into runs of text and http(s) URLs so the URLs can be
+ * rendered as links without interpreting the text as HTML.
+ */
+export const linkify = (text: string): TextPart[] => {
```
jennifer-richards, line 14: Linkifying text is something of a pain - is there not a suitable library that has already dealt with corner cases for us as an alternative to rolling our own?

client/app/pages/docs/[id]/editorial-notes.vue
```diff
@@ -0,0 +1,138 @@
...
+  { server: false, lazy: true }
+)
+
+const isEditing = ref(false)
+const isSaving = ref(false)
+const draftText = ref('')
+
+const mailArchiveHref = computed(() => {
+  const name = rfcToBe.value?.name
+  return name?.startsWith('draft-') ? mailArchiveSearchUrl(name) : undefined
```
Comment (jennifer-richards) [inline client/app/pages/docs/[id]/editorial-notes.vue:107]: curious: why `undefined` rather than `null`?

rpc/api.py
```diff
@@ -2735,6 +2737,40 @@ def post(self, request, format=None):
...
+        operation_id="documents_editorial_note_update",
+        request=EditorialNoteSerializer,
+        responses=EditorialNoteSerializer,
+        parameters=[OpenApiParameter("draft_name", OpenApiTypes.STR, "path")],
+    )
+    def put(self, request, draft_name: str):
+        rfc_to_be = resolve_rfctobe(draft_name)
+        note, _ = EditorialNote.objects.get_or_create(rfc_to_be=rfc_to_be)
+        serializer = EditorialNoteSerializer(note, data=request.data)
+        serializer.is_valid(raise_exception=True)
```
jennifer-richards, line 2771: possible nit: when first adding a note, this will leave behind an empty note even if validation fails. A transaction would, I believe, prevent that.

That's technically mostly harmless but might be confusing for a user who just hit an error and gets mixed messages about whether they updated the notes or not. (I haven't checked whether this is visible to the user - it may not be, but I suspect it will as the empty note having been updated by the `get_or_create()` call)

rpc/models.py
```diff
@@ -1133,6 +1133,26 @@ def __str__(self):
...
 
 
+class EditorialNote(models.Model):
+    """Shared free-text scratchpad the RPC keeps for an RfcToBe"""
+
+    rfc_to_be = models.OneToOneField(
+        RfcToBe, on_delete=models.CASCADE, related_name="editorial_note"
+    )
+    text = models.TextField(blank=True, default="")
+    updated_at = models.DateTimeField(auto_now=True)
```
jennifer-richards: Mixing `auto_now` with manual `updated_by` tracking may attribute changes incorrectly, but leave a misleading timestamp. Not sure whether that's better or worse than an unexpected update not touching the timestamp at all.

jennifer-richards (changes requested):
Couple nits / questions, a couple places the model creation / timestamp handling can leave behind misleading situations. It's probably not essential to fix the latter, but worth a look to see if there's an easy way.

client/app/pages/docs/[id]/editorial-notes.vue
Comment (rudimatz) [reply client/app/pages/docs/[id]/editorial-notes.vue:107]: could be both, but `undefined` consistent with types around it, e.g. `rfcToBe.value?.name` in line above
semantic could be seen as:
null -> not set 
undefined => not applicable (which is the case here)

rpc/api.py
rudimatz: will refactor to only save after successful validation

rpc/models.py
rudimatz: I'll change it to explicitly set the time with the API call - like this updated_by and updated_at go hand-in-hand

client/app/pages/docs/[id]/editorial-notes.vue
Comment (rudimatz) [reply client/app/pages/docs/[id]/editorial-notes.vue]: ok will remove

client/app/utils/linkify.ts
rudimatz: will use library to use on URLs

jennifer-richards (approved):

